PR #10, opened 2019-07-17
FIX: Engage Brakes on Limit Hits and Stop Button

ZLLentz:
Currently, the brakes fail to engage when a limit is hit or when the IOC asks for a stop. This resolves the issue.

The new pathway is:
limit hit -> bExecute goes to FALSE -> brake engages

Or:
"stop" button hit -> bExecute goes to FALSE -> brake engages


lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -69,8 +72,16 @@ IF stMotionStage.bError THEN
...
-IF stMotionStage.Axis.Status.InTargetPosition AND fbDriveVirtual.bDone THEN
+// Check the limits and cancel execution if appropriate. Without this block we have infinite error spam
+bFwdHit := stMotionStage.Axis.Status.PositiveDirection AND NOT stMotionStage.bLimAllFwd;
+bBwdHit := stMotionStage.Axis.Status.NegativeDirection AND NOT stMotionStage.bLimAllBwd;
+IF bFwdHit OR bBwdHit THEN
+	stMotionStage.bExecute := FALSE;
+END_IF
+
+// Mark done if settled or if unexpected stop, then reset the FB if there is no error
+fbExecFallingEdge(CLK:=stMotionStage.bExecute);
```
teddyrendahl: My only issue is that have an `if` statement that checks for the falling edge of `bExecute` and inside that if statement we also set `bExecute` to False. Obviously this will work, double setting the `bExecute` isn't a huge deal. The downsides are 1) That block will fire twice on a standard move (no error or limit) 2) The code is kind of confusing. I'll leave it up to you to decide if you want to break these into two separate statements somehow?

lcls-twincat-motion/Library/Library.plcproj
```diff
@@ -16,9 +16,9 @@
     <LibraryReferences>{9dd7de72-3863-4684-ae1f-9a4b8617466d}</LibraryReferences>
     <CombineIds>false</CombineIds>
     <Company>SLAC</Company>
-    <Released>true</Released>
+    <Released>false</Released>
     <Title>lcls-twincat-motion</Title>
-    <ProjectVersion>0.0.2</ProjectVersion>
+    <ProjectVersion>0.0.0</ProjectVersion>
```
klauer: Version regressed?

lcls-twincat-motion/Library/Library.tmc
```diff
@@ -1,7 +1,7 @@
...
...terOutputUpdate</Name><Type GUID="{18071995-0000-0000-0000-000000000030}">BOOL</Type><BitSize>8</BitSize><BitOffs>240</BitOffs></SubItem><SubItem><Name>RTViolation</Name><Type GUID="{18071995-0000-0000-0000-000000000030}">BOOL</Type><BitSize>8</BitSize><BitOffs>248</BitOffs></SubItem><SubItem><Name>TaskName</Name><Type GUID="{18071995-0000-0000-0000-00010000003F}">STRING(63)</Type><BitSize>512</BitSize><BitOffs>512</BitOffs></SubItem><Hides><Hide GUID="{6A76D020-03A2-465C-A678-C341951E9EF3}" /><Hide GUID="{6F7D679F-72A0-4831-A7F1-085F839743ED}" /></Hides></DataType></DataTypes><Modules><Module GUID="{DED1545C-5FCD-48E2-B686-7D2C250A3641}" TcSmClass="TComPlcObjDef"><Name>Library</Name><CLSID ClassFactory="TcPlc30">{08500001-0000-0000-F000-000000000064}</CLSID><Licenses /><Contexts /><Parameters /><DataAreas /><Deployment /><EventClasses /><Properties><Property><Name>ApplicationName</Name><Value>Port_851</Value></Property><Property><Name>ChangeDate</Name><Value>2019-07-16T17:41:15</Value></Property><Property><Name>GeneratedCodeSize</Name><Value>217088</Value></Property><Property><Name>GlobalDataSize</Name><Value>57344</Value></Property></Properties></Module></Modules></TcModuleClass>
```
klauer: Can't see what changed here effectively. I wonder if we should have a commit hook that runs an xml formatter on these files?

lcls-twincat-motion/Library/Version/Global_Version.TcGVL
```diff
@@ -8,7 +8,7 @@
 VAR_GLOBAL CONSTANT
 	{attribute 'const_non_replaced'}
 	{attribute 'linkalways'}
-	stLibVersion_lcls_twincat_motion : ST_LibVersion := (iMajor := 0, iMinor := 0, iBuild := 2, iRevision := 0, sVersion := '0.0.2');
+	stLibVersion_lcls_twincat_motion : ST_LibVersion := (iMajor := 0, iMinor := 0, iBuild := 0, iRevision := 0, sVersion := '0.0.0');
```
klauer: Ditto on version?

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -93,7 +105,7 @@ END_CASE
 // Handle the brake. TRUE means brake disabled.
```
klauer: Mostly unrelated to this PR, but if `.bBrake` `True` means disabled, think we could rename it to `.bBrakeDisengaged` or disabled... or something even less confusing?

lcls-twincat-motion/Library/Library.plcproj
ZLLentz: 0.0.0 is the code for "not a release version". If you install a project with a version number on the PLC programming node it overwrites any existing project with that number.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: I agree that this is confusing. I'll think a bit about if and how to revise.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: I agree, but I couldn't come up with a succinct name. `True` is disabled because all of our breaks are NC.

lcls-twincat-motion/Library/Library.tmc
ZLLentz: I'm not sure how to set such a thing up but it's a good idea.

lcls-twincat-motion/Library/Library.plcproj
klauer: Huh, that's... interesting... and well-worth documenting if we don't already have it somewhere.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: The simplest solution is to add something like "AND NOT bDoneResolved", setting bDoneResolved to TRUE at the end of this block and to FALSE somewhere earlier in the code, perhaps when we start a move

lcls-twincat-motion/Library/Library.plcproj
ZLLentz: I agree, I forgot about it when making this repo, which caused a problem, which is why it's bundled into here sloppily. The idea is to set versions to 0.0.0 and use https://github.com/pcdshub/tc_release when we want to apply a tag and set a version number.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -69,8 +72,16 @@ IF stMotionStage.bError THEN
 	stMotionStage.bEnable := FALSE;
 END_IF
 
-// Mark done if settled, then reset the FB if there is no error
-IF stMotionStage.Axis.Status.InTargetPosition AND fbDriveVirtual.bDone THEN
+// Check the limits and cancel execution if appropriate. Without this block we have infinite error spam
+bFwdHit := stMotionStage.Axis.Status.PositiveDirection AND NOT stMotionStage.bLimAllFwd;
```
slacAWallace: You're using this combination to account for any unforseen halt in a given direction?

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
slacAWallace: Pedantic, but I think it's helpful to really clarify disabled == disengaged.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
slacAWallace: > 1) That block will fire twice on a standard move (no error or limit)

How does this fire twice? I am looping trying to understand. You definitely don't want that to be happening.

ZLLentz:
Found only one issue with this today while testing the code with the PPM and XPIM. Both throw an error every time they hit the upper limit (but not the lower!!?...), error is: `The axis or coupled slave axis has lost its controller enable signal while executing a command (error 0x4260)`

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: Should I approach this by adding a huge comment, or by finding an appropriate rename?

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: Yes. If any kind of forward limit is triggered while moving forward, we note `bFwdHit` and engage the brake appropriately.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: If the block is entered normally (the move ends successfully with no interruptions), then the block ends by setting `bExecute` from `TRUE` to `FALSE`. This triggers the falling edge, satisfying the condition to enter the code block again. After the second pass, this doesn't happen again. No bad code is run by doing this, but I agree that it's bad practice and unclean.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
teddyrendahl: The crucial point is that `Status.PositiveDirection` has to be there to ensure we can move off of the limit 😉

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
klauer: My opinion: I think an appropriate rename would be advisable, or confusion in boolean state could easily lead to bugs down the line.

ZLLentz: Some options, keeping in mind that this is a hardware output variable:
```
bDisableBrake
bBrakeDisable
bNoBrake
bUnBrake
bReleaseBrake
bBrakeRelease
bBrakeOff
```

teddyrendahl:
`bBrakeRelease` gets my vote

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
klauer: I'll second @teddyrendahl .... `bBrakeRelease`

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: Actually, a super simple fix: just add `AND NOT stMotionStage.bDone` to the conditional

teddyrendahl:
Don't care about order just the combination of words 😄

ZLLentz:
I prefer `bBrakeRelease` so as your eye scans across the variable, it sees "Brake" first so you immediately know what it is, then "Release" serves to explain the polarity.

Are similar changes needed for the limits? Some may also be confused for `TRUE` meaning "limit not hit"

klauer:
Always think of variables in terms of tab completion `.bBr<tab>` ;)

It'd probably be advisable for limits to be tweaked as well.

ZLLentz:
Ok, I will open a separate API break PR for renaming stuff

ZLLentz:
I can reliably reproduce the error in a simpler program by setting `bEnable` to `FALSE` on the same cycle as `bLimFwd` is set to `FALSE`. This indicates to me that somewhere in this code I'm killing `bEnable` before the `MC_Power` function block as a chance to register the limit as hit. I bet I can remove the error just by rearranging the sections a bit.

ZLLentz:
Some further testing shows that it takes 5 cycles before we are allowed to disable the motor without erroring. I'm not quite sure how I should resolve it.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -69,21 +72,34 @@ IF stMotionStage.bError THEN
...
+		stMotionStage.bDone := TRUE;
+		stMotionStage.bBusy := FALSE;
+		IF NOT stMotionStage.Axis.Status.Error THEN
+			bExecute := FALSE;
+			stMotionStage.bExecute := FALSE;
 		END_IF
 	END_IF
 END_IF
+IF stMotionStage.nEnableMode = ENUM_StageEnableMode.STAGE_ENABLE_MOTION AND NOT bExecute THEN
+	IF stMotionStage.Axis.Status.MotionState = MC_AXISSTATE_STANDSTILL THEN
```
ZLLentz: This conditional is what resolves the extra errors issue

ZLLentz:
I am happy with this PR now. If there is some agreement, I'll merge and then double-back for the renaming PR.

PR #11, opened 2019-07-18
ENH: Add function block and DUT for p2p state moves

ZLLentz:
Maggie and I worked through this, but we still need to test it. We realized that we definitely need a function block for triggering moves to specified states. The idea is that the user can put to a boolean PV to go to the position and see relevant information. @ghalym

Later we could feed the information for these from some sort of PMPS databse.


ZLLentz:
Added pragmas, etc., going to do some superficial testing now

ZLLentz:
Ok I ended up moving tons of stuff around because it was getting annoying to test.. but this is confirmed working. I suspect the API may need to be tweaked to play nicely with PMPS but it is fine for now.

ZLLentz:
Also, I'm merging this is as-is for a quick tag/build so the cc library stuff can move forward. Anything that looks awful can be addressed in a follow-up PR.

PR #14, opened 2019-07-23
API: Rename several confusing vars

### LINKED ISSUE #12: API Changes Needed for Clarity
Rename bBrake -> bBrakeRelease to make the polarity of the boolean crystal clear.
Similar changes are also needed for bLimFwd and bLimBwd.

See discussion in #10

Proposed renames:
`bBrake` -> `bBrakeRelease`
`bEnabled` -> `bEnableDone`
`bLimFwd` -> `bLimFwdOk`
`bLimBwd` -> `bLimBwdOk`

ZLLentz:
Accepting suggestions for:
1. Better names
2. Other vars to permanently rename before we are locked in

List of renames:
- `bBrake` -> `bBrakeReleased` to indicate polarity
- `bEnabled` -> `bEnableDone` to remove visual confusion with `bEnable` and to make clear the intent of the variable
- `bLimFwd` -> `bLimFwdOk` (and equivalent change on all limit variables) to indicate polarity and status as a permission variable

closes #12

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
@@ -28,7 +28,7 @@ CASE stMotionStage.nEnableMode OF
 		IF stMotionStage.bExecute THEN
 			stMotionStage.bEnable := TRUE;
 		END_IF
-		bExecute := stMotionStage.bExecute AND stMotionStage.bEnabled;
+		bExecute := stMotionStage.bExecute AND stMotionStage.bEnableDone;
 END_CASE
 
 // Automatically fill the correct nCmdData for homing
@@ -101,15 +101,15 @@ END_IF
 CASE stMotionStage.Axis.Status.MotionState OF
 	// We are not enabled if there is an issue
 	MC_AXISSTATE_UNDEFINED, MC_AXISSTATE_DISABLED, MC_AXISSTATE_ERRORSTOP:
-		stMotionStage.bEnabled := FALSE;
+		stMotionStage.bEnableDone := FALSE;
 	ELSE
-		stMotionStage.bEnabled := TRUE;
+		stMotionStage.bEnableDone := TRUE;
 END_CASE
 
 // Handle the brake. TRUE means brake disabled.
 CASE stMotionStage.nBrakeMode OF
 	ENUM_StageBrakeMode.STAGE_BRAKE_DISABLED:
-		stMotionStage.bBrakeRelease := stMotionStage.bEnabled AND stMotionStage.bExecute;
+		stMotionStage.bBrakeRelease := stMotionStage.bEnableDone AND stMotionStage.bExecute;
 	ENUM_StageBrakeMode.STAGE_BRAKE_STANDSTILL:
 		IF stMotionStage.Axis.Status.MotionState = MC_AXISSTATE_STANDSTILL THEN
 			stMotionStage.bBrakeRelease := FALSE;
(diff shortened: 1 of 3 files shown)


teddyrendahl:
`bLimFwdOk` still seems vague. However a better alternative escapes me....

slacAWallace:
`bLimFwdEnb`

klauer:
Is there a reason we need to keep things so... brief?

`bLimitForwardEnabled`? `bLimitForwardOK` `bLimitMoveForwardOK` `bForwardMotionAllowed` `bForwardEnableEPSSummary` `bForwardEnable`

`bForwardGantryEnable`

It seems like as soon as you put 'limit' in the name, the logic gets confusing. `Limit = True` in English means the opposite of `Enable = True` - so `bLimitEnable`...

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -141,86 +141,8 @@ IF stMotionStage.nHomingMode = ENUM_EpicsHomeCmd.EPICS_MOTOR_HOME_NONE THEN
...
-      <LineId Id="162" Count="0" />
-      <LineId Id="158" Count="0" />
-      <LineId Id="155" Count="0" />
-      <LineId Id="143" Count="4" />
-      <LineId Id="9" Count="0" />
-      <LineId Id="174" Count="1" />
-      <LineId Id="102" Count="0" />
-      <LineId Id="176" Count="0" />
-      <LineId Id="104" Count="5" />
+      <LineId Id="404" Count="123" />
```
klauer: Unrelated, but I'm curious - is this block describing the visual studio folds?

ZLLentz: I have no idea what these blocks describe or why they all disappeared after using the built-in find and replace mechanism

ZLLentz:
No, we don't need to keep the names brief (though I'd avoid using super verbose names). How about `b<thing><direction>Enable`, with `<thing>` picking from `Limit`, `All`, Gantry`, and `<direction>` picking from `Forward` and `Backward`.

ZLLentz:
Also, if `Limit` is confusing, we can use the name `Switch` instead

klauer:
I think that's a reasonable suggestion. I'm probably thinking too much/being too pedantic about the `Limit` thing, so feel free to ignore me on that one.

ZLLentz:
I'll assume everyone is either happy with those proposed names or too busy to put up an argument against them, so I'll commit with the above guidelines with the intention of merging tomorrow.

I don't want to change the API twice, so please let me know concerns. It's already kind of annoying that I waited this long to do it.

klauer:
I’d say go for it. :+1:

ZLLentz:
A quick explanation on `bEnableDone`:
The function block sets `bEnable` to `TRUE`, which does not immediately enable the motor, so we have to wait for the `MC_Power` block to respond. This usually happens within a few cycles, but could theoretically take longer if we're going through something like PMPS to ensure the beam state is ok. `bEnableDone` flips to `TRUE` once the enabling process is done, so our code knows that it is ok to start moving the motor.

As I type this out, I realize that we may need to extend this to waiting on the `Enable_<direction>` bits being set, depending on the interface we agree upon with PMPS.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
slacAWallace: I think it's folds. Given how many temporary files any ide makes, you'd think they could track it some other way.

slacAWallace:
bEnable to bEnabled vs bEnableDone. Between the last two, the latter is more distinct, even if it's a bit awkward. I tend to past-tense stuff, but Done makes it stand out better.

ZLLentz:
Ok, there's no going back now. Closing the door on API changes forever.

PR #15, opened 2019-08-05
ENH: Gantry Library Auto-Coupling Function Block

jsheppard95:
This is a pull request to begin discussion/obtain feedback on gantry-specific function blocks in `lcls-twincat-motion`.

**Summary of Changes:**
1. Added `bSTO` input variable in `DUT_MotionStage` and tied into `FB_DriveVirtual.bEnable`
2. Added logic to make gantried axes respect each other's limits - i.e when a master axis is coupled its EnableFwd/Bwd bits are taken as its own AND that of the slave axis
3. Added logic to reset coupling status bit `bGantryAlreadyCoupled` along with couple start/stop bools `bExecuteCouple` and `bExecuteDecouple`

**Usage in HOMS project:**
Project using library: https://github.com/jsheppard95/HOMS_Test/tree/gantry-library
Summary:
`// Start Coupling/Decoupling Logic`
`fbAutoCoupleX : FB_GantryAutoCoupling;`

`fbAutoCoupleX(fGantryTol:=5000, Master:=M3, MasterEnc:=x_up_enc, Slave:=M4, SlaveEnc:=x_dwn_enc, bExecuteCouple:=bExecuteCoupleX, bExecuteDecouple:=bExecuteDecoupleX, bGantryAlreadyCoupled=>bGantryAlreadyCoupledX);`

`fbMotionStage_x_up  (stMotionStage:=M3);`
`fbMotionStage_x_dwn (stMotionStage:=M4);`

And this would repeat for the Y gantry motors. The operator can then couple/decouple the motors by writing `TRUE` to `bExecuteCoupleX` and `bExecuteDecoupleX`.

**Automated Testing Brainstorming:**
I meant this pull request as a way to solicit feedback before moving too far, and the grey areas I'm seeing so far are:
1. What needs to be tested and where should this occur (i.e in library or HOMS specific code)?
  a) Tests to confirm coupling starts & stops as expected upon user input - likely in library
  b) Tests to confirm coupled response to limit switches - likely in HOMS code
  c) Tests to confirm logic of bFwd/Bwd enables as a result of too great a gantry difference - likely in library
  d) Additional tests?


jsheppard95:
Just realized maybe the first thing I should do is update everything to the new lcls-twincat-motion API before the branches keep fragmenting

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
```diff
@@ -0,0 +1,89 @@
...
+	Master.bLimGntFwd := Master.bLimGntFwd AND Slave.bLimGntFwd AND Slave.bLimFwd;
+	Slave.bLimGntFwd := Master.bLimGntFwd AND Slave.bLimGntFwd AND Master.bLimFwd;
+	// Bwd limits
+	Master.bLimGntBwd := Master.bLimGntBwd AND Slave.bLimGntBwd AND Slave.bLimBwd;
+	Slave.bLimGntBwd := Master.bLimGntBwd AND Slave.bLimGntBwd AND Master.bLimBwd;
+END_IF
+
+// Coupling states
+// Auto-coupling at init and auto-reset of coupling boolean
+bExecuteCouple S= NOT bInitComplete;
```
ZLLentz: I started reviewing but got stuck on these `S=` and `R=` operators. What do these operators do?

jsheppard95: Oh yes Alex showed those to me. The command `S=` reads as "Set by" so for example `A S= B` means when B goes to True A also goes to True but not vice-versa. `R=` reads as "Reset by" and does the opposite so `A R= B` means when B goes to true A goes back to false.

So the logic here is:
1) `bExecuteCouple S= NOT bInitComplete;`
2) `bExecuteCouple R= couple.Busy OR bGantryAlreadyCoupled;`
3) `couple(Master:=Master.Axis, Slave:=Slave.Axis, Execute:=bExecuteCouple);`
4) `bInitComplete S= bGantryAlreadyCoupled;`
5) `bExecuteDecouple R= decouple.Busy OR NOT bGantryAlreadyCoupled;`
6) `decouple(Slave:=Slave.Axis, Execute:=bExecuteDecouple);`

Since all bools start False, on startup (1) makes `bExecuteCouple = True` which makes the `couple` function block execute. Then once coupling is confirmed by `bGantryAlreadyCoupled`, `bInitComplete` will go to `True` which causes the `couple` FB to not be repeatedly executed. Then when the user wants to decouple they write `bExecuteDecouple` to `True` which makes line (6) execute, and then this bool resets back to `False` by line (5) after the axes are no longer coupled

ZLLentz: Ok, I understand now, thanks. These are missing from Beckhoff's documentation on structured text operators, but they've left two rows for them. See https://infosys.beckhoff.com/english.php?content=../content/1033/tcplccontrol/html/tcplcctrl_iec_operators_overview.htm&id=2174038239638951926, two blank boxes where there should be `S=` and `R=`.

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
```diff
@@ -0,0 +1,89 @@
...
+bGantryAlreadyCoupled := Master.Axis.NcToPlc.CoupleState=1 AND Slave.Axis.NcToPlc.CoupleState=3;
+
+// Make master and slave axes respect each other's limits
+IF bGantryAlreadyCoupled THEN
+	// Fwd limits
+	Master.bLimGntFwd := Master.bLimGntFwd AND Slave.bLimGntFwd AND Slave.bLimFwd;
+	Slave.bLimGntFwd := Master.bLimGntFwd AND Slave.bLimGntFwd AND Master.bLimFwd;
+	// Bwd limits
+	Master.bLimGntBwd := Master.bLimGntBwd AND Slave.bLimGntBwd AND Slave.bLimBwd;
+	Slave.bLimGntBwd := Master.bLimGntBwd AND Slave.bLimGntBwd AND Master.bLimBwd;
```
ZLLentz: Is there a reason why we can't use the more complete `AND Slave.bAllFowardEnable`, etc. variables? Are we not guaranteed that `FB_MotionStage` is running for the slave axis?

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
```diff
@@ -0,0 +1,89 @@
...
+END_IF
+
+// Coupling states
+// Auto-coupling at init and auto-reset of coupling boolean
+bExecuteCouple S= NOT bInitComplete;
+
+bExecuteCouple R= couple.Busy OR bGantryAlreadyCoupled;
```
ZLLentz: What happens if `couple.Busy` is `TRUE` and we reset this, then enter the active `couple` function block with the `Execute` arg set to `FALSE`? Does this stop the coupling in progress? Or does it do nothing? Or does this never happen because the coupling happens in one cycle?

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -42,7 +42,7 @@ stMotionStage.bLimAllBwd := stMotionStage.bLimBwd AND (stMotionStage.bLimGntBwd
 
 // Handle standard commands using ESS's FB
 fbDriveVirtual(En:=TRUE,
-	bEnable:=stMotionStage.bEnable,
+	bEnable:=stMotionStage.bEnable AND stMotionStage.bSTO,
```
ZLLentz: We need to refactor this a bit to allow an external function block with custody of `MC_Power` to correctly enable/disable without needing to learn how to correctly handle `bSTO`, like what we did with the limits. Also, we may consider renaming `bSTO` to make it more obvious that `bSTO = TRUE` is what is needed to allow motion.

ZLLentz:
I forgot to say that I like this PR overall :+1: despite having comments everywhere

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
klauer: Those S/R operators are interesting. I wonder if it means "set" and "reset" as in an "SR flip-flop" - or more verbosely in this case "positive edge-triggered set and reset", I suppose.

lcls-twincat-motion/Library/DUTs/DUT_MotionStage.TcDUT
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<TcPlcObject Version="1.1.0.1" ProductVersion="3.1.4022.12">
+<TcPlcObject Version="1.1.0.1" ProductVersion="3.1.4020.14">
```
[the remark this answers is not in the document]
klauer: Yes! 4022 all the way. cc @slacAWallace

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
```diff
@@ -0,0 +1,89 @@
...
+    <Implementation>
+      <ST><![CDATA[// Designate Master and SLave Axes
+Master.bGantryAxis := TRUE;
+Slave.bGantryAxis := TRUE;
+
+Master.fGantryTol := fGantryTol;
+Slave.fGantryTol := Master.fGantryTol;
+
+// Activate Gantry Virtual Limit Switch
+gantry_diff_limit(Penc:=MasterEnc, SEnc:=SlaveEnc, GantDiffTol:=LREAL_TO_LINT(Master.fGantryTol),
```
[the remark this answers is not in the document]
klauer: :+1: for encoder-count tolerance - it's less ambiguous.

lcls-twincat-motion/Library/DUTs/DUT_MotionStage.TcDUT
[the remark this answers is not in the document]
ZLLentz: Yes! But we also need to pick a 4022.XX version, or else we'll trade back and forth every PR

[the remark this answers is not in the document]
klauer: Ooh, good point. I don't know anything about the 4022`.XX` version differences.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
klauer: Reading the variable name as-is: `"boolean Safe-Torque-Off = True" means Enable`

:+1: on the rename.

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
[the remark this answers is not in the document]
slacAWallace: good point, this really should be in counts.

slacAWallace:
I would love to see this pr come in with some tests, or another pr shortly after with some tests.

ZLLentz:
It's hard for me to ask someone to PR with tests when the core library doesn't have any `tcunit` stuff yet

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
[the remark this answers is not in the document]
jsheppard95: Makes sense to me, so we'll change `DUT_MotionStage.fGantryTol to` to be `LINT`, rename to be `.nGantryTol`, then take out the conversion?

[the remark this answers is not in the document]
ZLLentz: Yeah, I think that's the logical change.

jsheppard95:
Yes I think my next step is going to be to write some basic tests, maybe just confirming coupling and decoupling upon writing to the corresponding variables to start. That way as I start to implement these changes it will be much quicker to verify things are still working.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz: `bHardwareEnable`?

I just realized that all existing stages without hardware STO buttons will need to lock this variable to `TRUE`, but I guess that's ok? Also anything with a redundant STO won't be able to link to this variable directly, e.g. `M1.bSTO := bSTO1 AND bSTO2`, but I guess that's ok too

klauer: `bHardwareEnable` sounds fine.

@ZLLentz I think that's just a detail that users will have to work around.

klauer:
I'm really excited to see TcUnit in use (maybe not as much as @slacAWallace) but maybe we do a couple fixes, merge this, and move on with the next steps?

jsheppard95:
@klauer That's fair, I've been working on a TcUnit for this function block but the only successful test I've written is ensuring coupling on startup, other tests such as ensuring motion stops when the gantry gets too large have been more difficult for me to construct. Maybe for now I'll make another PR with the one working test I have then I'll start making these changes and return to the tests later.

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
jsheppard95: Now that you say it I don't have a reason why we aren't using the `bAllForwardEnable` variables. I believe we are guaranteed that the slave still uses `FB_MotionStage`, and this seems like the same logic but would be much easier to read. I'll try now and confirm

jsheppard95: I'm actually not so sure about this. The main issue that led to this code snippet is the fact that if a Slave's limit switch activates before a master's, motion will continue for both axes. I tried the following:
```
IF bGantryAlreadyCoupled THEN
	Master.bLimAllFwd := Master.bLimAllFwd AND Slave.bLimAllFwd;
	Master.bLimAllBwd := Master.bLimAllBwd AND Slave.bLimAllBwd;
	
	Slave.bLimAllFwd := Master.bLimAllFwd;
	Slave.bLimAllBwd := Master.bLimAllBwd;
END_IF
```
But this gives me the same "master ignoring slaves limit switch issue", so I do think the `AND Slave.bLimFwd` lines are necessary. If that's true, then doing something like:
```
IF bGantryAlreadyCoupled THEN
	Master.bLimAllFwd := Master.bLimAllFwd AND Slave.bLimAllFwd AND Slave.bLimFwd;
	Master.bLimAllBwd := Master.bLimAllBwd AND Slave.bLimAllBwd AND Slave.bLimBwd;
	
	Slave.bLimAllFwd := Master.bLimAllFwd;
	Slave.bLimAllBwd := Master.bLimAllBwd;
END_IF
```
is somewhat confusing because the `bLimAllFwd` variables are themselves set by the normal limits.

Maybe I'm not seeing the right approach, but given all of this it seems simpler to have this code snippet only set `LimGntFwd` variables and let `FB_MotionStage` set `bLimAllFwd` accordingly.

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
jsheppard95: So I'm not entirely sure and trying to figure out a way to step through the cycles and test this but here are my initial thoughts:
1. Coupling occurs on the second cycle, so what you're saying could potentially happen.
2. If we enter the active `couple` function block with `bExecuteCouple = FALSE`, it won't stop coupling if it has already occurred, that was why I added the `decouple` function block to provide this functionality.
3. Since coupling didn't execute on this cycle, it seems to me that `bInitComplete` would remain False which then would set `bExecuteCouple` to True on the next cycle and it would just take an extra cycle to complete coupling.

I need to try and step through this though as I could be overlooking something

jsheppard95: I just tried this by setting breakpoints from the `bExecuteCouple S= ` line until the end and I don't think this will be an issue. I found that `couple.Busy` stays False until after the `couple` function block's execution begins and then goes to True. Even if someone were to reset this to False before coupling finishes, this doesn't stop the `couple` function block that's currently getting executed, and then after it completes `couple.Busy` goes to False anyway so the desired logic continues executing.

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
ZLLentz: I meant to use the `LimAll` variables on the right hand side instead of expressions like `Slave.bLimGntFwd AND Slave.bLimFwd`

jsheppard95:
So I'm unsure why I needed to force push, as commit 4fde9fc is not present in my local history but instead is 2732809 which is also present in the history on GitHub. Regardless the only comment I did not address in my latest commit is that referring to `Slave.bAllFowardEnable`. The way I implemented that change gave me some strange logic that I'm currently trying to understand/fix. I'll comment in that thread with more updates.

ZLLentz:
@jsheppard95 you always need to force push after a rebase since you're rewriting the commit tree

lcls-twincat-motion/Library/POUs/Motion/FB_GantryAutoCoupling.TcPOU
jsheppard95: So my latest commit caaba7e implements this by setting `bAllForwardEnable` (and Backward/AllEnable) in its own function block. This allows these variables to be set in both `FB_MotionStage` and `FB_GantryAutoCoupling` which was necessary to avoid logic loops where `bAllForwardEnable` was stuck at `FALSE`. @slacAWallace originally suggested using a function as opposed to a function block, but this gets tricky as functions have only one output, meaning I would really need three functions, one of which would set `bAllForwardEnable`, another `bAllBackwardEnable`, and the third `bAllEnable`. I did read that functions can have multiple outputs but its unclear if these can be attributes of a structure.

Really my question though is why should this not be a function block as I'm not immediately seeing the downside of this.

ZLLentz:
I'm planning to do some git surgery on this to remove the misc header changes and then merge it in as-is. We can do the other changes we want (testing, small refactors) in follow-up PRs.

PR #16, opened 2019-08-29
ENH: Multi state FBs

ZLLentz:
The goal is to have something that maps nicely onto the old EPICS IOCs and have a consistent way of doing these multi-state things so the IOCs all look the same. This isn't priority to merge or to get into any projects, but I'd like to move state devices in a more uniform direction.

I'm not planning to merge this yet, but I'm putting up the PR.


ZLLentz:
closed in favor of #31

PR #17, opened 2019-09-03
MNT: Some cleanup and reorganization

ZLLentz:
Quickly moving things around e.g. putting gantry stuff in it's own folder, etc.


ZLLentz:
Dropping the `.tmc` was not intentional. I suspect it will come back next time the library is built...

PR #22, opened 2019-10-11
MNT: Add docstrings and pragmas to DUT_MotionStage

ZLLentz:
Added a pytmc pragma to every variable. Moved and edited the docstrings
for clarity and consistency, and so that TwinCAT will pick them up when
viewing the struct online.

### DIFF
--- a/lcls-twincat-motion/Library/DUTs/DUT_MotionStage.TcDUT
+++ b/lcls-twincat-motion/Library/DUTs/DUT_MotionStage.TcDUT
@@ -9,31 +9,31 @@ STRUCT
 	Axis: AXIS_REF;
 	// NC Forward Limit Switch: TRUE if ok to move
 	{attribute 'pytmc': '
-		pv: IO:ENA:FWD
+		pv: PLC:bLimitForwardEnable
 		io: i
 	'}
 	bLimitForwardEnable AT %I*: BOOL;
 	// NC Backward Limit Switch: TRUE if ok to move
 	{attribute 'pytmc': '
-		pv: IO:ENA:BWD
+		pv: PLC:bLimitBackwardEnable
 		io: i
 	'}
 	bLimitBackwardEnable AT %I*: BOOL;
 	// NO Home Switch: TRUE if at home
 	{attribute 'pytmc': '
-		pv: IO:HOME
+		pv: PLC:bHome
 		io: i
 	'}
 	bHome AT %I*: BOOL;
 	// NC Brake Output: TRUE to release brake
 	{attribute 'pytmc': '
-		pv: IO:BRAKE
+		pv: PLC:bBrakeRelease
 		io: i
 	'}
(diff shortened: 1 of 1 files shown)


ZLLentz:
Accepting all opinions about what the PV names should actually be. A lot of these may be useful for debugging later.

lcls-twincat-motion/Library/DUTs/DUT_MotionStage.TcDUT
```diff
@@ -1,59 +1,224 @@
...
     <Declaration><![CDATA[TYPE DUT_MotionStage :
 STRUCT
-	// Hardware
+	(* Hardware *)
+	
+	// PLC Axis Reference
 	Axis: AXIS_REF;
+	// NC Forward Limit Switch: TRUE if ok to move
+	{attribute 'pytmc': '
+		pv: IO:ENA:FWD
```
klauer: Just an observation: the final suffix is uppercase in these, but camelcase in the settings section.

lcls-twincat-motion/Library/DUTs/DUT_MotionStage.TcDUT
```diff
@@ -1,59 +1,224 @@
...
+		pv: GANTRY:ENA:BWD
+		io: i
+	'}
+	bGantryBackwardEnable: BOOL:=FALSE;
 	
-	// Settings
+	(* Settings *)
+	// Change to FALSE if MC_POWER is somewhere else in the project
+	{attribute 'pytmc': '
+		pv: SETTING:bPowerSelf
```
klauer: I like that these reflect the actual variable names in the PLC program.

ZLLentz: I'm wondering if I should go back through and match them all up to the PLC variables with no effort put in to make "normal looking" PVs. In some sense these are all "debug" PVs that don't need to be user-facing in any capacity, so making it obvious where they match up inside the PLC is a positive. In my first run-through I was imagining using some of these in a user-facing way (not that users see PV names... usually)

klauer: Generalizing like this sounds good to me: 
* Categorize with a prefix (`Setting:`, for example, here)
* Use the variable name as-is for the final portion

ZLLentz: I agree, I will change it as you describe.

ZLLentz:
Quick conclusion before I merge this: I settled on the prefix `PLC:` for all of these because it makes it very clear where the values are coming from. These PVs are all intended for debug purposes.

An example PV in the new motion system would be something like `IM3L0:PPM:MMS:Y:PLC:bBrakeRelease_RBV`

Most of these are read-only, but I left a few writable because they may be useful debug points.

PR #24, opened 2019-11-16
API: State PVs and Args

ZLLentz:
Simplify state move permission args into a single `bMoveOk`, redo all the PVs to make them cleaner and more consistent.


ZLLentz:
One other thing I'd like to note here if someone stumbles upon this later:

I decided to make the pragmas all IO, despite us at times not wanting all of them to be writable from EPICS, because having a structure that is sometimes missing PVs makes interfacing difficult on the EPICS side for ophyd and for screens. Later we will find a way to either implement immutable PVs on the PLC side or mark them as non-writable in EPICS in a cohesive way.

PR #27, opened 2019-11-23
FIX: remove pragmas from LINTs

ZLLentz:


klauer:
We could _technically_ bring `LINT`s in with pytmc, but I worry the lossy conversion could lead to confusion:

https://github.com/pcdshub/twincat-ads/blob/24f81b23c52355ea0088bebd509680470e5f24b5/adsApp/src/adsAsynPortDriver.cpp#L3597

ZLLentz:
I think it's ok to require the TwinCAT code to decide how to bring the LINTs down to 32 bit.

PR #29, opened 2019-12-12
ENH: Add a basic Visu for DUT_MotionStage

ZLLentz:
This isn't mature yet, but I've started a twincat visu for DUT_MotionStage. It may be useful for assembly-area checkouts in the future or for local debugging. I'm not going to merge this into the main repo for a while, instead I'm going to iterate the visu inside my PPM and REFL test projects and update back to here occasionally until I think it is sufficiently polished. Feel free to try out this build if you're interested.


ZLLentz:
Also, it's possible that it'd be better to put the Visu into its own Library to avoid interfering with normal operation with library loading/compiling issues, etc., not 100% sure though.

klauer:
I saw this on your computer, but a screenshot here would be nice. 😄 👍

ZLLentz:
image

ZLLentz:
closing in favor of #106 which is closer to being what we want

PR #38, opened 2020-02-15
STY: No tabs or trailing whitespace

ZLLentz:
There was an issue in my last PR where some files have tabs, some files have spaces, and some files have trailing whitespaces. I'd like to address this now before moving forward. I expect that if refined a bit this can be used in other twincat projects as well.

Added `style/check_style.sh` to be run in travis
Added `style/fix_style.sh` to fix the style locally

I'm pushing this first without doing `fix_style.sh` to demonstrate that it fails. Then, I will fix the style, push again, and demonstrate that it succeeds.

closes #38

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,18 @@ matrix:
     - name: Documentation building
       python: 3.7
       env: TWINCAT_BUILD_DOCS=1
+    - name: Twincat Style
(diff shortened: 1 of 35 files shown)


ZLLentz:
Forcing the merge through, going back to fix the new pytmc pragma in a separate PR

PR #45, opened 2020-02-21
FIX: pragma typos

ZLLentz:
oh `pymtc`


ZLLentz:
Is there any reasonable way to check for this in the linter? I suspect maybe not...

klauer:
Fuzzy check the attribute name? 🤷‍♂

PR #57, opened 2020-03-11
Set up Example IOC

ZLLentz:
Super basic IOC setup from the Example project

Contains 3 simulated axes:

1. Works and moves everywhere
2. Refuses to move always
3. Refuses to move beyond some virtual limit switches near 0 and 100

Needs some extra work to be truly useful but I think it's good enough to roll with for now


klauer:
Looking good so far! 👍

ZLLentz:
@klauer I don't plan on updating this PR, but I plan on follow-ups to make this more useful for testing `ads-ioc`. Let me know if you have ideas.

klauer:
OK. The most pressing thing I can think of would be verifying the archive setting functionality that was added.

Beyond that, it'd be neat to have a full-stack example: motion example PLC project + happi configuration + screens and such... Hmm...

ZLLentz:
Full stack example is a good call. I'll create a couple issues and then merge this PR.

PR #67, opened 2020-04-28
MNT: CI, Pre-commit, Example Cleanup

### LINKED ISSUE #63: Move Example IOC to own Project/Repo
There are a lot of things about the Example IOC that are slightly non-standard because it exists as the second PLC project inside what is primarily a Library-oriented solution. There's also a lot of cross-talk where the projects both build when you try to build one of them. Using the Example IOC as a starting point for a test IOC is cumbersome due to the current Axis Id restrictions on the driver. It would be better if it had its own project/repo.

ZLLentz:
- Update CI to latest best practices
- Add pre-commit hooks
- Move Example PLC project to its own repo, for ease of use and to make it a more representative example: https://github.com/pcdshub/lcls-twincat-motion-example
- Set target PLC to local instead of the motion ioc testing PLC
- Update `typhos` test screens for recent renaming
- Remove unused `Config` directory. This had some experimental utilities for hooking up to NC and CoE values. This can be revived via the magic of version control when it is needed again. I'll make sure there is an issue about this so it is not forgotten.

closes #63


ZLLentz:
It helps when almost 20,000 lines are just formatting a tmc file

PR #72, opened 2020-06-17
WIP: Unit Tests

ZLLentz:
Make draft PR so I don't forget that I was working on this prior to going back to finishing the FEE


ZLLentz:
This has fallen so far off the radar that I'd like to take it down and bring it back next year with a fresh take

PR #91, opened 2020-09-15
ENH: MPS Upgrade

ZLLentz:


lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
```diff
@@ -12,9 +12,16 @@ VAR_IN_OUT
     fbArbiter: FB_Arbiter;
     fbFFHWO: FB_HardwareFFOutput;
 END_VAR
+VAR_INPUT
+    stFullBeam: ST_BeamParams := PMPS_GVL.cstFullBeam;
```
ghalym, line 16: can we call it like Maximum Allowed Beam? Just so we don't confuse it with FullBeam.

ghalym (approved):

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
ZLLentz: I agree with a rename, but that name is not quite right either. We're actually defining a minimum beam setting at which a state is configured safe enough for the FFO to be automatically reset, not a maximum allowed beam.

slacAWallace: PossibleMaxBeam? MaxBeamPossible?

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
```diff
@@ -54,32 +57,51 @@ END_FOR]]></ST>
...
+ffZeroRate(
+    i_xOk := stBeamNeeded.nRate > 0,
+    i_xAutoReset := TRUE,
+    i_DevName := stMotionStage.sName,
+    i_Desc := 'Device requesting zero rate',
+    i_TypeCode := 16#1001,
+    io_fbFFHWO := fbFFHWO);
+
+// Trip the beam for BPTM timeouts if we want to move
+ffBPTMTimeoutAndMove.i_xReset S= bAtSafeState;
```
slacAWallace, line 95: I forget does this reset internally?

slacAWallace (changes requested):
Looks like what we discussed.

I would suggest double checking the logic of the resets they may provide more rising edges that you'd like.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
ZLLentz: I do not understand the question.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
ZLLentz: `stMaxBeamThreshold` is what I'm going with

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
ZLLentz: Some investigation shows that i_xReset is reset at the HWO end, not at FB_FastFault, so this needs to be adjusted a bit

PR #92, opened 2020-09-17
MNT: fix issues related to BPTM usage

ZLLentz:
- Always reset stage errors from motion request FBs
- fix errors wrt telling the BPTM motion is done, and listening for BPTM
  done state
- hold safety ready low instead of high when not doing state transitions

Together, these fixes make the state motion with PMPS very very reliable.


ZLLentz:
Slightly longer summary:
I made a lot of the logic more explicit and I expect this to work every single time now.

PR #99, opened 2020-10-14
MNT: move arbiter pragma so EPICS works

### LINKED ISSUE #94: PMPS Arb enable PV doesn't work
bArbiterEnable is pointing to the wrong level in the FB tree, so writes are overwritten by the PLC

ZLLentz:
The pragma was putting to a lower-level caller that gets overridden in a higher level. This resolves the issue so that the built IOC will set/get from the correct variable.

closes #94


ZLLentz:
Merging simple bugfix after self-review

PR #101, opened 2020-10-14
MNT: fix accidentally changed PV

ZLLentz:
PMPS:ARB:ENABLE became ARB:ENABLE, oops


ZLLentz:
self-merging simple bugfix

PR #106, opened 2020-10-24
ENH: add motion stage visu from zllentz/tc3-checkout

klauer:
(Largely untested and I may have screwed something up based on the line diff count; so don't merge)

When you have visualizations in the library, they apparently show up in projects that use it:
image

* Adds DUT_MotionStageExtras (required for visualization; should maybe consider merging it with DUT_MotionStage?)
* Adds VISU_MotionStage

### DIFF
--- a/lcls-twincat-motion/Library/DUTs/DUT_MotionStage_Extras.TcDUT
+++ b/lcls-twincat-motion/Library/DUTs/DUT_MotionStage_Extras.TcDUT
@@ -1,0 +1,13 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<TcPlcObject Version="1.1.0.1" ProductVersion="3.1.4024.4">
+  <DUT Name="DUT_MotionStage_Extras" Id="{1e6627e8-279b-4e38-87a2-ae79853e08df}">
+    <Declaration><![CDATA[TYPE DUT_MotionStage_Extras :
+STRUCT
+    fVisuStep: LREAL;
+    bVisuLowLim: BOOL;
+    bVisuHighLim: BOOL;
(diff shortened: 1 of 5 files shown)


ZLLentz:
maybe compare with #29 on Monday

klauer:
Completely forgot about that PR, sorry @ZLLentz!

Likely will close this one in favor of #29, but keeping it here for reference in the meantime.

ZLLentz:
This one is probably basically the same thing but without the merge conflicts

klauer:
TODO: reference https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=295101434 in docs

klauer:
@vanxiong has his own visualization that does not require lcls-twincat-motion as well:
image

ZLLentz:
That looks pretty snazzy
I'd like to include both probably

ZLLentz:
So the only reasons I don't want to immediately merge this is because:
- I want to merge the "extras" struct into the main DUT
- I don't know how library visualization dependencies work: if we include this, does it mean that all projects that depend on this need to load the visu libraries to function properly?

I'm going to start from this branch and test a few things

ZLLentz:
Ok so this doesn't load on prog node with 4022.30
TwinCAT vizu configuration is pain

klauer:
It would be unfortunate if every sub-project pulled in motion's visualization requirements unnecessarily.
I had a bunch of version-related issues when trying to deploy this. I think I had made the visualizations in 4024 and then failed going back to 4022.30. Not sure if this PR was before/after that.

klauer:
Summarizing in #148 and closing this for now

PR #110, opened 2020-11-14
MNT: bump MAX_STATES to 9

klauer:
As we don't have good support for pytmc pragmas that allow for partial inclusion of structures, I believe this is the quickest solution to support AT1K4, with its 8 filter states and 1 out state (and the implied unknown state).

Ref https://github.com/pcdshub/lcls-plc-tmo-motion/pull/39


ZLLentz (approved):
I think we merge this without thinking too hard about it. There is a chance this could cause an issue in the LFE motion IOC where it will take forever to boot with having so many extra PVs... but it's not really a big deal.

PR #118, opened 2021-01-07
MNT: add waits where appropriate in the homing routines

### LINKED ISSUE #113: Homing set position can fail if axis does not stop in a timely manner
Try waiting for the axis to fully stop before doing set position

ZLLentz:
closes #113 

Simply wait for the axis to stop prior to setting the position to home, to avoid the homing failure where we try to set the position of an axis still in motion.

I reproduced the bug on my test stage, then routed all transitions in the state machine that should wait through a new waiting state. Apply the patch to my test stage, then test the homing- all was well.

This has previously affected the scatter slits and the XTES imager zoom/focus, but this should fix both cases.


ZLLentz:
Confirmed that this fixes the XTES lens motor homing with Phil

ZLLentz:
Merging without review because it seems to work

PR #122, opened 2021-01-26
FIX: Always reset command PVs

### LINKED ISSUE #121: Homing Race Condition
If the variable to start a homing routine (`bHomeCmd`) is held high for multiple PLC cycles, it will get stuck high because the code only automatically resets this variable on rising edge. This will prevent the user from issuing another homing command with no way to reconcile.

ZLLentz:
closes #121 

Sometimes, the IOC may send "set the cmd to True" twice via screen use- not sure if this is double-clicking with perfect timing, some sort of write queuing, or something else. This breaks the session because I was only resetting the homing command on rising edge.

This maintains the "start homing on rising edge" behavior, but it allows the command to reset and be re-issued even if this double-put happens.


klauer (approved):
Good find and good fix. 

(Mostly to self: the additional layers of typhos/pcdsdevices with how `DUT_MotionStage.bHomeCmd` gets set certainly adds to the mental load required to run through the logic. Need to add this to my mental list of things to check when reviewing similar edge-triggered things.)

ZLLentz:
Yes, I had the same thought Ken: one more thing to keep in mind when setting up epics-edge-triggered PLC variables... I assume the motor record is doing some smarts for us here.

PR #125, opened 2021-02-11
ENH: fix issues with interrupting state moves

### LINKED ISSUE #123: Investigate States FFO Error State from 12:10 AM Feb 3rd
https://github.com/pcdshub/lcls-twincat-motion/blob/25f33e83c451b406af8e883a2a7f12388b52743c/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU#L91

IM4L0 was faulting because it was between states. This FFO was tripped, indicating potential logic issues- where was the accompanying "in-motion" fast fault?

This "stuck between states" state should always give us a good FFO message like "hey, this device is stuck between states" rather than "Device requesting zero rate"

Ideally we figure out why the problem happened at all to see if we can prevent it from happening again.

ZLLentz:
There are at least ~2~ ~3~ 4 bugs right now related to #123 

This fixes all the known bugs (closes #123)

- If the axis is already moving when we ask for a move, wait for the previous move to truly end before starting again. Otherwise, the new move command just hangs forever.
- Do not start a move if we plan to disable the axis in the next cycle. Also stop a move in this scenario. Otherwise, this throws a confusing error instead of just... stopping.
- Reset missing variables in the bReset of PMPS state moves. Otherwise, these variables can get stuck in strange states after errors or when interrupting moves.
- Do not fully reset the state FB when we interrupt it with a new move, this actually breaks everything and causes confusing errors.

Also:
- Fix testing bug in the pmps tester fb

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
@@ -116,7 +116,7 @@ fbSetEnables(stMotionStage:=stMotionStage);
 
 // Do not start the move until we have power and the safety system says it is OK
 IF bMoveCmd THEN
-    bExecute := stMotionStage.bExecute AND stMotionStage.bAllEnable AND stMotionStage.bEnableDone AND stMotionStage.bSafetyReady;
+    bExecute := stMotionStage.bExecute AND stMotionStage.bAllEnable AND stMotionStage.bSafetyReady AND stMotionStage.bEnableDone AND stMotionStage.bSafetyReady;
 ELSE
     bExecute := stMotionStage.bExecute;
 END_IF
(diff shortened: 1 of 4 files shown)


ZLLentz:
Some of this is actually a little messy... there could be more work here to clean up some things. But in this state it 100% works in my offline testing. I will need to test again with the real arbiter during PAMM.

klauer (approved):
I don't think I'll get the opportunity to give you a proper review before you really might be able to put it to use. For what it's worth, the diff passes the smell test.

I'll still try to give it a more thorough read-through tomorrow, in any case, but merge now by all means - based on your successful local testing.

klauer:
OK, I gave this and the surrounding context a much more thorough read-through. The PR contents look good, fixing what I'd consider pretty subtle bugs. I made some notes about future documentation, but really have nothing further to add that would have any actual effect on the issues at hand.

TLDR: LGTM. 👍

ZLLentz:
Thanks Ken. The more I revisit this code the more I think it needs some big refactoring... but I'm not sure how long it would take or even how to do it properly to make problems like this more obvious.

slacAWallace (commented):
I didn't get a moment to say I looked at the PR today and that the changes look like improvements simply because I can understand them better than what was there before.

👍

PR #128, opened 2021-02-24
MNT: fix issue where auto-enable block could disable

ZLLentz:
Follow-up to #125 
We check `bSafetyReady` on the start of a new move request to see if we should enable the motor yet.
However, if we were in the middle of an active move, then get a new move request, this block of code can actually disable the enabled motor, causing an error.
This PR changes the "moving" blocks here to only be able to enable the motor, not disable it.
This is the last piece needed for #123 

The homing enable is untouched because I don't want to test it during a PAMM.

There is other code later in this function block intended to handle the motor disabling.

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
@@ -90,7 +90,7 @@ CASE stMotionStage.nEnableMode OF
                 stMotionStage.bEnable := stMotionStage.bSafetyReady;
             ELSIF bPosGoal THEN
                 IF stMotionStage.bAllForwardEnable THEN
-                    stMotionStage.bEnable := stMotionStage.bSafetyReady;
+                    stMotionStage.bEnable S= stMotionStage.bSafetyReady;
                 ELSIF NOT stMotionStage.bError THEN
                     // Not an error, just a warning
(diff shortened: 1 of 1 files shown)


klauer (approved):
Stressful to find _during_ a PAMM - but good work on debugging and fixing this 👍

ZLLentz:
Tested via mashing different options on the states combobox on im1l1, verifying in twincat scope that things stay enabled, things switch directions and that the brake isn't toggling

ZLLentz:
image

slacAWallace:
Why is bSafety false at the start of a new move? Waiting on things like BPTM to auth a new transition?

ZLLentz:
Yes, exactly

ZLLentz:
This is the correct behavior, the function block just needs to be smart enough not to cut power to the motor while it's trying to stop normally.

ZLLentz:
This is also why I did not see this behavior offline

ZLLentz:
ok let's deploy and hopefully move on to not-pamm-things

slacAWallace:
That leaves something to be desired... this is a have your cake and eat it too situation. You can't have 100% assured BPTM functionality and on-the-fly interruptible moves. Even with everything operating at light speed. This fix makes sense, but it makes me sad, and want to think more deeply about the nature of the BPTM.

slacAWallace (approved):

PR #133, opened 2021-03-25
MNT: Various Race Condition Fixes

### LINKED ISSUE #130: State moves with arbiter timeout get stuck in permanent fast fault
Moving to any state with an arbiter timeout fast fault causes the fault to be latched permanently

The fault is supposed to be reset when `bInTransition` goes `FALSE` or when the goal state resets to 0, but neither of these things ever happen somehow

ZLLentz:
closes #130


ghalym (approved):

ZLLentz:
I know this is hard to review, I'm going to claim that I've fully tested it and that it works. We might need to revisit later, hopefully not.

PR #139, opened 2021-10-07
BUG: Save/restore race condition

ZLLentz:
Under testing conditions (small project, boot into config mode, no auto-start boot project), the save/restore works great.

Under prod conditions (large project, boot into run mode, auto-start boot project), it's possible for it to fail.
My hypothesis is that the plc code starts running before all of the ADS/NC infrastructure is ready, and if the load is high then this can result in a race condition where the `MC_SetPosition` fails.

This PR simply adds 10 retries at 1 second intervals. This absolutely overkill, in testing this always worked on the second time if it failed on the first, but it seemed appropriate.

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/FB_EncSaveRestore.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/FB_EncSaveRestore.TcPOU
@@ -65,14 +65,14 @@ END_VAR
             bWaitRetry := TRUE;
         END_IF
     END_IF
-    
+
     tonRetry(
         IN := bWaitRetry,
         PT := tRetryDelay);
-        
+
     bLoad S= tonRetry.Q;
     bWaitRetry R= tonRetry.Q;
-    
+
     // Check DUT_MotionStage for an encoder error (range 0x44nn)
     bEncError := stMotionStage.bError AND stMotionStage.nErrorId >= 16#4400 AND stMotionStage.nErrorId <= 16#44FF;
 


lcls-twincat-motion/Library/POUs/Motion/FB_EncSaveRestore.TcPOU
```diff
@@ -45,17 +49,35 @@ END_VAR
...
-        // Alert the user that something has gone wrong
-        stMotionStage.bError := TRUE;
-        stMotionStage.nErrorId := fbSetPos.ErrorID;
-        stMotionStage.sCustomErrorMessage := 'Error loading previously saved position.';
+        nCurrTries := nCurrTries + 1;
+        IF nCurrTries >= nMaxRetries THEN
+            // Alert the user that something has gone wrong
+            stMotionStage.bError := TRUE;
+            stMotionStage.nErrorId := nLatchError;
+            stMotionStage.sCustomErrorMessage := 'Error loading previously saved position.';
```
klauer, line 57: I wonder if this is worth a centralized log message so we can determine how prevalent this is. I'm leaning toward "no" as the patch seems likely sufficient.

ZLLentz: I'm 50/50 on this, I expect this to fix everything but I also would prefer to notice and fix it sooner next time if it breaks again somehow.

Taking a step back, should all stMotionStage errors become a centralized log message? Well, that might be redundant if the error is also an NC error. There's time to think, I won't deploy this fix anywhere until the PAMM.

klauer: That's a good point, and I agree that it would be appropriate.

It reminds me of what @ghalym experienced the other day with ADS-based polling being insufficient to pick up on an error that only occurred during a single PLC cycle, whereas the asynchronously-emitted log message made its way through without issue.

And that reminds me - we might want to set all error messages such as this to 'monitor' instead of poll on the pragma side...

ZLLentz: https://github.com/pcdshub/lcls-twincat-motion/pull/140

PR #141, opened 2021-10-21
ENH: add a motor power kill switch

ZLLentz:
- Add a pragma-d `bKill` element to `DUT_MotionStage`
- If `bKill` is `TRUE`, disable the motor

This allows someone to kill the motor power entirely in the event that something bad has happened, provided that the IOC is online.


ZLLentz:
Hah, looks like I already did this ages ago and forgot! Funny stuff.
https://github.com/pcdshub/lcls-twincat-motion/pull/132

PR #147, opened 2022-02-14
ENH: add option for auto clearing beam parameter mismatch fast faults

nrwslac:
## Description

`FB_PositionStatePMPS(FB)` function block currently doesn't allow for  `i_xAutoReset` to be set to `True` causing all position state PMPS motion systems to require a manual reset. Here we add an input variable `bAutoReset` that defaults to `False`. PMPS states can now be implemented as FFO auto resetting.

## Motivation and Context
Users have to currently auto clear faults on MR1L0 and MR2L0 to permit beam.

## How Has This Been Tested?
Not tested yet.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
@@ -14,7 +14,7 @@ VAR_IN_OUT
 END_VAR
 VAR_INPUT
     stHighBeamThreshold: ST_BeamParams := PMPS_GVL.cstFullBeam;
-    bAutoReset: BOOL := False;
+    bBPOKAutoReset: BOOL := False;
 END_VAR
 VAR
     bptm: BeamParameterTransitionManager;
@@ -93,10 +93,7 @@ bAtSafeState := F_SafeBPCompare(stHighBeamThreshold, stBeamNeeded);
 ffBeamParamsOk.i_xOK := bFFOxOk;
 ffBeamParamsOk.i_xReset S= bFFOxOk AND bAtSafeState;
 ffBeamParamsOk.i_xReset R= NOT ffBeamParamsOk.i_xOK;
-
-IF bAutoReset THEN
-    ffBeamParamsOk.i_xAutoReset := True;
-END_IF
+ffBeamParamsOk.i_xAutoReset := bBPOKAutoReset;
 
 ffBeamParamsOk(
     i_DevName:=stMotionStage.sName,
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
@@ -14,6 +14,7 @@ VAR_IN_OUT
 END_VAR
 VAR_INPUT
     stHighBeamThreshold: ST_BeamParams := PMPS_GVL.cstFullBeam;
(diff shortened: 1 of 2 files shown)


lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
```diff
@@ -14,6 +14,7 @@ VAR_IN_OUT
 END_VAR
 VAR_INPUT
     stHighBeamThreshold: ST_BeamParams := PMPS_GVL.cstFullBeam;
+    bAutoReset: BOOL := False;
```
ZLLentz, line 17: This isn't a good name for the variable, since it implies that the entire set can be auto-reset. I don't know if this is the best either but it's more clear the intent.
```suggestion
    bBPOKAutoReset: BOOL := False;
```

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
```diff
@@ -92,6 +93,11 @@ bAtSafeState := F_SafeBPCompare(stHighBeamThreshold, stBeamNeeded);
 ffBeamParamsOk.i_xOK := bFFOxOk;
 ffBeamParamsOk.i_xReset S= bFFOxOk AND bAtSafeState;
 ffBeamParamsOk.i_xReset R= NOT ffBeamParamsOk.i_xOK;
+
+IF bAutoReset THEN
+    ffBeamParamsOk.i_xAutoReset := True;
+END_IF
```
ZLLentz, line 99: This is strangely indirect, why not something like this?
```suggestion
ffBeamParamsOk.i_xAutoReset := bAutoReset;
```

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
nrwslac: making this specific to beam parameters, makes sense.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
nrwslac: right, this an unnecessary if statement.

ZLLentz:
The automatic CI testing failure is not related to this PR
The change here is fully backwards-compatible so it looks pretty safe to try it out- worst case scenario is that it doesn't help, but I think it will.
How should we proceed- merge? Install as dev library for a test and update the optics to set this to TRUE?
We aren't poking at the behavior of the PMPS internals so it's not clear how much the PMPS simulator is needed here

nrwslac:
> The automatic CI testing failure is not related to this PR The change here is fully backwards-compatible so it looks pretty safe to try it out- worst case scenario is that it doesn't help, but I think it will. How should we proceed- merge? Install as dev library for a test and update the optics to set this to TRUE? We aren't poking at the behavior of the PMPS internals so it's not clear how much the PMPS simulator is needed here

How should we simulate this if we have the dev library up? We could force the FFO low on a mirror and see if it resets?

nrwslac:
Do FFO faults that don't affect operations kill the beam? example, sxr mirror fast fault while hxr taking beam.

ZLLentz:
Cross-line faults should be fine (faulting XRT during TMO-only time, for example). I don't remember if the stopper needs to be in or not to properly veto it.

ZLLentz:
Debugged and working for homs MR1L3! Merging for tag

PR #151, opened 2022-09-12
DEV: Adding the option to auto reset FFO for pneumatic actuators.

ghalym:
## Description
Adding the option to auto reset FFO for pneumatic actuators.

## Motivation and Context
ST1K4 in TMO is a photon stopper that has a pneumatic actuator. The FFO is latching, so when the actuator is retracted, operators still need to go to the PMPS screen to Ack the fault. This way, when instantiating the function block and setting the i_xAutoreset bit to True, they don't need to Ack the fault from the PMPS diagnostics screen.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/FB_MotionPneumaticActuator.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/FB_MotionPneumaticActuator.TcPOU
@@ -13,7 +13,7 @@ VAR_INPUT
     ibCntrlHold:BOOL; (* Control Signal must retain its value, must be TRUE in the case of single acting*)
     ibOverrideInterlock:BOOL; (*if true interlocks are ignored*)
     // Reset fault
-	{attribute 'pytmc' := '
+    {attribute 'pytmc' := '
     pv: FFO_Reset
     '}
     i_xReset: BOOL;
@@ -157,7 +157,7 @@ stPneumaticActuator.bError R= stPneumaticActuator.bReset;
 
 (*FAST FAULT*)
 fbFF(i_xOK := xMPS_OK,
-	i_xReset := i_xReset,
+    i_xReset := i_xReset,
     i_xAutoReset := i_xAutoReset,
     io_fbFFHWO := io_fbFFHWO);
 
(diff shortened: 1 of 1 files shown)


lcls-twincat-motion/Library/POUs/Motion/FB_MotionPneumaticActuator.TcPOU
```diff
@@ -14,9 +14,9 @@ VAR_INPUT
     ibOverrideInterlock:BOOL; (*if true interlocks are ignored*)
     // Reset fault
     {attribute 'pytmc' := '
-    pv: FF_Reset
+    pv: FFO_AutoReset
     '}
-    i_xReset: BOOL;
+    i_xAutoReset: BOOL;
```
ZLLentz, line 19: Is it possible to include this option while also keeping the old `i_xReset` API, rather than straight replacing it? This ends up renaming a PV and renaming an argument which could be annoying for the other places where we use this FB.

Here's a search that shows all of our uses of this FB:
https://github.com/search?q=org%3Apcdshub+FB_MotionPneumaticActuator&type=code

Is isn't that many but all of these would need to be edited on update

ghalym: Good point! Thank you. I put the original API back in.

ZLLentz:
I think this is :+1:

I'm going to to do the small style tweaks that travis wants and then hit merge

lcls-twincat-motion/Library/POUs/Motion/FB_MotionPneumaticActuator.TcPOU
```diff
@@ -13,10 +13,14 @@ VAR_INPUT
     ibCntrlHold:BOOL; (* Control Signal must retain its value, must be TRUE in the case of single acting*)
     ibOverrideInterlock:BOOL; (*if true interlocks are ignored*)
     // Reset fault
-    {attribute 'pytmc' := '
-    pv: FF_Reset
+	{attribute 'pytmc' := '
```
ZLLentz, line 16:
```suggestion
    {attribute 'pytmc' := '
```

lcls-twincat-motion/Library/POUs/Motion/FB_MotionPneumaticActuator.TcPOU
```diff
@@ -153,7 +157,8 @@ stPneumaticActuator.bError R= stPneumaticActuator.bReset;
 
 (*FAST FAULT*)
 fbFF(i_xOK := xMPS_OK,
-    i_xReset := i_xReset,
+	i_xReset := i_xReset,
```
ZLLentz, line 160:
```suggestion
    i_xReset := i_xReset,
```

ZLLentz (approved):

PR #158, opened 2023-01-21
FIX: Fixes for soft-line deployment

ZLLentz:
## Description

1. Remove the deprecated MAX_STATES backcompat.
2. Allow a pmps state mover to swap out and transition to new pre-emptive beam states without a physical move, so long as the new beam state has a unique ID.
3. Fixed the inverted logic for the unknown state FFO.
4. Fix a bug where database reads would be immediately overwritten instead of used.
5. Add pytmc pragmas to let us inspect which specific parameters have been loaded for a state mover.
6. Style updates
7. Automatic build changes in Global_Version.TcGVL

## Motivation and Context

1. The MAX_STATES backcompat was causing compilation problems on projects that did not explicitly refer to `lcls-twincat-general`.
2. Soon, we will need to implement "WFS" mode for certain imagers, which decreases the required beam level needed. This should happen in-place when the WFS goes in without needing to move the imager.
3. The logic was exactly backwards as implemented in the previous PR, oops!
4. This was preventing newer versions of the code from loading from the database
5. This lets the client transfer tool inspect live parameters
6. These either annoyed me or were required to make CI pass
7. This happened automatically and I don't really care to change it back

## How Has This Been Tested?

Interactively using plc-kfe-motion and plc-kfe-gatt

This is mostly fixes that will be documented in the release notes.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
@@ -26,6 +26,7 @@ VAR
     nTransitionAssertionID: UDINT;
     fbReadPmpsDb: FB_JsonDocToSafeBP;
     ftDbBusy: F_TRIG;
+    rtReadDBExec: R_TRIG;
     ftRead: F_TRIG;
     bptm: BeamParameterTransitionManager;
     ffBeamParamsOk: FB_FastFault;
@@ -174,10 +175,13 @@ IF ftDbBusy.Q THEN
     bReadPmpsDb S= NOT MOTION_GVL.fbPmpsFileReader.bError;
 END_IF
 
-arrPMPS[0].sPmpsState := sTransitionKey;
-FOR nBPIndex := 1 TO GeneralConstants.MAX_STATES BY 1 DO
-    arrPMPS[nBPIndex] := arrStates[nBPIndex].stPMPS;
-END_FOR
+rtReadDBExec(CLK:=bReadPmpsDb);
+IF rtReadDBExec.Q THEN
+    arrPMPS[0].sPmpsState := sTransitionKey;
+    FOR nBPIndex := 1 TO GeneralConstants.MAX_STATES BY 1 DO
+        arrPMPS[nBPIndex] := arrStates[nBPIndex].stPMPS;
+    END_FOR
+END_IF
 
 fbReadPmpsDb(
     bExecute:=bReadPmpsDb,
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
@@ -164,7 +164,7 @@ ffMaint(
 
 // Trip the beam for unknown state
 ffUnknown(
-    i_xOK := getState <= 0 AND NOT bInTransition,
+    i_xOK := getState > 0 OR bInTransition,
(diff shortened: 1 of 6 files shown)


ZLLentz:
I want to merge this and deploy a new tag for use. There might end up being one or two more bugs but things look promising as-is.

ZLLentz:
TODO:
- [ ] Clean up PR text
- [ ] Get the CI to pass
- [ ] Ask for reviews

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
```diff
@@ -155,7 +164,7 @@ ffMaint(
 
 // Trip the beam for unknown state
 ffUnknown(
-    i_xOK := getState <= 0 AND NOT bInTransition,
+    i_xOK := getState > 0 OR bInTransition,
```
klauer, line 167: Friday afternoon is making logic harder than it should be. Would you mind explaining the logic inversion change here to put my confusion to rest?

`getState > 0` means in known state
`bInTransition` means arbiter beam parameter transition in progress

```
OK = getState > 0 OR bIntransition
Not OK = not (getState > 0 OR bIntransition)
Not OK = getState <= 0 AND NOT bIntransition
```

Not OK if in an unknown state _specifically when not in a BPTM transition_
OK if in a known state or in a BTPM transition

klauer (approved):
I'm clearly no expert on the latest PMPS changes required here, but in isolation, the individual changes seem reasonable to me.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU
ZLLentz: This will be a bit rambley

In the PR description I should have put: "fixed the inverted logic for the unknown state FFO" instead of "Add a fast fault for the case where we've stopped moving between states.", oops.

The logic from last PR was exactly inverted. I forgot that the API was "ok => no fault" and I had tried to put in "fault if state is unknown and not doing a bptm transition", instead the pre-diff code reads "things are ok only if the state is unknown and we are not doing a bptm transition". As a result, we had a permenant fast fault when at known states.

Your description of the current behavior is correct: I want a fault in the cases where our state is unknown. That is, if we start up the PLC and a device is already between states, that's a fault, and if a device stops between states, that's also a fault.

Previously, this was taken care of by putting in an "Unknown" state assertion to the arbiter at "no beam".
Now, with database-generated states, this would require an additional "Unknown" database entry for every device, since we need to have unique IDs for the arbiter to know that it's a different set of beam parameters. So, rather than put the burden on the database, do the check here: enforce no beam via fast fault.

ZLLentz: And further, I don't want a fault in transition, I want to retain the option for a "transition state" to have nonzero beam parameters for cases where that's OK.

klauer: Thanks for the explanation - that does clear it all up 👍

ZLLentz: Thank you for making me explain/justify/document the reasoning, it was definitely missing

ZLLentz:
I'm going to merge this and put out a tag so we can assemble projects against a tag build instead of dev.
It's reasonably likely there's 1-2 bugs left but this might also be a stable version for a while.

PR #159, opened 2023-01-26
BUG: inset -> insert

nrwslac:
## Description

`DUT_MotionPneumaticActuator` has a miss spelled pragma for the PV: bInsertDigitalOutput

typo, fix, insert 'r'

## Motivation and Context

devices can't connect with this typo as the PV will be wrong.

https://github.com/pcdshub/pcdsdevices/pull/1088

## How Has This Been Tested?

This typo was found while testing `pneumatic.py` 

The pneumatic device can access the PV with the typo.

## Pre-merge checklist


nrwslac:
Are we supposed to load this lib with 4024? Should I test this by saving the library to a dev version and setting a plc-project to dev motion?

ZLLentz:
Don't worry about the version header, it was probably a mistake to load it in 4024 in the first place. It won't cause any issues regardless.

ZLLentz (approved):
This is a good catch.

I think a typo fix is simple enough to merge without further testing, but if you want an IOC with the new value we can ship out a dev 0.0.0 build.

PR #164, opened 2023-03-30
Add VAR_OUTPUT triggered after a read of MOTION_GVL.fbPmpsFileReader

nrwslac:
## Description

- Add `VAR_OUT` `BOOL` `bReadPmpsDb` to `FB_Standard_PMPSDB`
- Set it high for once when `fbPmpsFileReader.bBusy` is finished.
- Creates a handle to trigger a reload of pmps database parameters not in state implementation

## Motivation and Context

https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=368318178
Asks for hard coded beam parameters to be loaded from the JSON doc.

## How Has This Been Tested?
This was tested using lcls-plc-tmo-optics/pull/35
https://github.com/pcdshub/lcls-plc-tmo-optics/pull/35
https://github.com/pcdshub/lcls-twincat-optics/pull/18

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
@@ -4,7 +4,6 @@
     <Declaration><![CDATA[FUNCTION_BLOCK FB_Standard_PMPSDB
 VAR_IN_OUT
     io_fbFFHWO: FB_HardwareFFOutput;
-    bRefreshBPs: BOOL;
 END_VAR
 VAR_INPUT
     // If TRUE, FB will run. Reads when enable goes TRUE.
@@ -26,6 +25,7 @@ VAR_OUTPUT
         io: i
     '}
     nLastRefreshTime: DINT;
+    bReadPmpsDb: BOOL;
 END_VAR
 VAR
     bExecute: BOOL;
@@ -44,9 +44,6 @@ END_VAR
     RETURN;
 END_IF
 
-IF bRefresh = TRUE THEN
-    bRefreshBPs := TRUE;
-END_IF
 rtEnable(CLK:=bEnable);
 rtRefresh(CLK:=bRefresh);
 bRefresh := FALSE;
@@ -69,6 +66,12 @@ MOTION_GVL.fbPmpsFileReader(
 );
 
 ftBusy(CLK:=MOTION_GVL.fbPmpsFileReader.bBusy);
+IF ftBusy.Q THEN
+    bReadPmpsDb := TRUE;
+ELSE
+    bReadPmpsDb := FALSE;
+END_IF
+
 
 // Lifted from Arbiter PLCs: keep track of the time
 fbTime(sNetID:='');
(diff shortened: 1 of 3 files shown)


ZLLentz:
- This isn't backwards-compatible with existing code.
- Why is this better then checking for the existing busy flag like is done in the states FBs?

https://github.com/pcdshub/lcls-twincat-motion/blob/20c1a5ba2e721335d58b23b07ac2dac39f5588c0/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU#L181-L184

ZLLentz:
Perhaps this could instead be implemented as a `VAR_OUTPUT` (for backwards compatibility) and feed it via looking for the busy transition as above (instead of holding high forever but only on refresh and not on startup)

nrwslac:
> * This isn't backwards-compatible with existing code.
> * Why is this better than checking for the existing busy flag like is done in the states FBs?
> 
> https://github.com/pcdshub/lcls-twincat-motion/blob/20c1a5ba2e721335d58b23b07ac2dac39f5588c0/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_PositionStatePMPS.TcPOU#L181-L184

I am reading this as, whenever the `fbPmpsFileReader` is busy, goes `TRUE`, wait for a it to finish Falling Trigger, then prompt a reload of the database. I haven't quite unpacked why `ReadPmpsDb` needs to latch TRUE

ZLLentz:
> I haven't quite unpacked why bReadPmpsDb needs to latch TRUE

In the linked code, that's used as an execute bit, so dropping it to FALSE would stop the execution. It goes to FALSE later when the execution is done.

In this PR, the bool is held TRUE forever, meaning it can only be triggered once.

nrwslac:
> > I haven't quite unpacked why bReadPmpsDb needs to latch TRUE
> 
> In the linked code, that's used as an execute bit, so dropping it to FALSE would stop the execution. It goes to FALSE later when the execution is done.
> 
> In this PR, the bool is held TRUE forever, meaning it can only be triggered once.


I see now there's a better way to do this with triggers, gonna fix the dependability of the caller having to do this, but as it is, since its a reference to a `BOOL`, its doesnt stay high if the caller sets it back to `FALSE`

ZLLentz (approved):
This version is useful, simple, re-usable, and backwards-compatible

ZLLentz:
@nrwslac do you want me to make and deploy a tag with this change?

nrwslac:
> @nrwslac do you want me to make and deploy a tag with this change?

yes please.

ZLLentz:
Ok here we go

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
```diff
@@ -64,6 +66,12 @@ MOTION_GVL.fbPmpsFileReader(
 );
 
 ftBusy(CLK:=MOTION_GVL.fbPmpsFileReader.bBusy);
+IF ftBusy.Q THEN
+    bReadPmpsDb := TRUE;
+ELSE
+    bReadPmpsDb := FALSE;
+END_IF
```
ZLLentz, line 73: Uhh I don't want to take back my merge, but next time note that you can do this in one line:
```
bReadPmpsDb := ftBusy.Q
```

nrwslac: Ah yes, that is cleaner. Thanks for pointing that out. Will use this next time.

ZLLentz:
This is now deployed on the programming node as feature release v2.2.0 :+1:

PR #184, opened 2023-07-28
WIP: state input helper

### LINKED ISSUE #176: Add helper functions for setting up state positions
## Current Behavior

No such helpers exist, and setting state position specifications manually is verbose and error-prone.

## Expected Behavior

We should create helpers that help us set up `ST_PositionState` individual instances and full state arrays.

The single struct version should let us define some defaults once and then apply them to individual position state instances, giving the user an easy way to say e.g. "each of these has a delta of 0.5, a velo of 1, (etc.)" and then set the positions and pmps keys individually.

The array version should be similar but could also let us express set positions in an array (?)

## Context / environment

We want to make the process of setting up states easier and less error-prone.

ZLLentz:
## Description

Provide a function block for making it less verbose and less error prone to set up large arrays of position states.

## Motivation and Context

Setting up position states is verbose and error-prone.
closes #176 

## How Has This Been Tested?

TODO: add unit test

TODO: include in readme

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
@@ -1,0 +1,136 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<TcPlcObject Version="1.1.0.1">
+  <POU Name="FB_StateSetupHelper" Id="{802f49f0-571f-413b-bf0f-1cf473b320d9}" SpecialFunc="None">
+    <Declaration><![CDATA[FUNCTION_BLOCK FB_StateSetupHelper
+(*
+    This is a helper for setting up large numbers of ST_PositionState instances.
+
+    This is typically difficult to do by hand in the normal ways and can be error-prone.
+    For each field, we will pick values with the following priority:
+    1. The value used in the function block call.
+    2. The value used in the most recent SetDefaults method call.
+    3. The default value as defined on ST_PositionState
+
+    Example expected usage:
+    VAR
+        fbStateSetup: FB_StateSetupHelper;
+        astStates1: ARRAY[1..GeneralConstants.MAX_STATES] OF ST_PositionState;
+        astStates2: ARRAY[1..GeneralConstants.MAX_STATES] OF ST_PositionState;
+        astStates2: ARRAY[1..GeneralConstants.MAX_STATES] OF ST_PositionState;
+    END_VAR
+
+    fbStateSetup.setDefaults(
+        fDelta := 0.5,
+        fVelocity := 10,
+        bMoveOk := TRUE,
+        bValid := TRUE,
+    );
+    fbStateSetup(stPositionState:=astStates1[1], sName:='OUT', fPosition:=10);
+    fbStateSetup(stPositionState:=astStates1[2], sName:='YAG', fPosition:=20);
+    fbStateSetup(stPositionState:=astStates1[3], sName:='TT', fPosition:=30);
+    fbStateSetup(stPositionState:=astStates2[1], sName:='OUT', fPosition:=-30);
+    fbStateSetup(stPositionState:=astStates2[2], sName:='YAG', fPosition:=35);
+    fbStateSetup(stPositionState:=astStates2[3], sName:='TT', fPosition:=70);
+    fbStateSetup(stPositionState:=astStates3[1], sName:='OUT', fPosition:=0.4, fDelta:=0.1);
+    fbStateSetup(stPositionState:=astStates3[2], sName:='YAG', fPosition:=2.3, fDelta:=0.1);
+    fbStateSetup(stPositionState:=astStates3[3], sName:='TT', fPosition:=5.6, fDelta:=0.1;
+*)
+VAR_IN_OUT
+    stPositionState: ST_PositionState;
+END_VAR
+VAR_INPUT
+    sName: STRING;
+    fPosition: LREAL;
+    nEncoderCount: UDINT;
+    fDelta: LREAL;
+    fVelocity: LREAL;
+    fAccel: LREAL;
+    fDecel: LREAL;
+    bMoveOk: BOOL;
+    bLocked: BOOL;
+    bValid: BOOL;
+    bUseRawCounts: BOOL;
+    sPmpsState: STRING;
+END_VAR
+VAR
+    stDefault: ST_PositionState;
+    stOriginal: ST_PositionState;
+END_VAR
+]]></Declaration>
+    <Implementation>
+      <ST><![CDATA[
+stPositionState.sName := sName;
+stPositionState.fPosition := fPosition;
+stPositionState.nEncoderCount := nEncoderCount;
+stPositionState.fDelta := fDelta;
+stPositionState.fVelocity := fVelocity;
+stPositionState.fAccel := fAccel;
+stPositionState.fDecel := fDecel;
+stPositionState.bMoveOk := bMoveOk;
+stPositionState.bLocked := bLocked;
+stPositionState.bValid := bValid;
+stPositionState.bUseRawCounts := bUseRawCounts;
+stPositionState.stPMPS.sPmpsState := sPmpsState;
+
+// Reset the inputs
+_InternalDefaultInit();
+]]></ST>
+    </Implementation>
+    <Action Name="_InternalDefaultInit" Id="{d6adfd4e-d299-4073-be2d-d70d90a830ad}">
+      <Implementation>
+        <ST><![CDATA[
+// Write to the input args so that unset args are the defaults
+sName := stDefault.sName;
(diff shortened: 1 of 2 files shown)


ZLLentz:
I'm very happy with the API I settled on here following the discussions on the linked issue. This remains a draft because the code is completely untested.

ZLLentz:
A bit of a snag here: method calls in TwinCAT3 cannot have default arguments.
Next week I'll find some other way to make this intuitive to use.

klauer (approved):
I think this is a big improvement! I really like the terse specification of states.

I don't have anything significant to add - the API looks good, the implementation looks good, and the tests look reasonably exhaustive.

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
```diff
@@ -0,0 +1,107 @@
...
+    stPositionState.fAccel := fAccel;
+    stPositionState.fDecel := fDecel;
+    stPositionState.bMoveOk := bMoveOk;
+    stPositionState.bLocked := bLocked;
+    stPositionState.bValid := bValid;
+    stPositionState.bUseRawCounts := bUseRawCounts;
+    stPositionState.stPMPS.sPmpsState := sPmpsState;
+END_IF
+
+// Overwrite the input args so that unset args are the defaults in the next call
```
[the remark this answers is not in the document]
ZLLentz: If the user forgets or decides not call `bSetDefault` this will use the "default defaults" as defined by the `ST_PositionState` struct, which are mostly `0`/`FALSE`/`''` (except the default name is `'Invalid'`).

Before doing something, I need to ask a quick question: are there any situations where we don't want to set a default? I think the answer is "no", you always want to at least opt in to `bValid:=TRUE`. So I agree that we should do something here.

I think the most prudent response is to make a warning message? Should I also force `bValid:=FALSE` so the state can't be used in the normal ways?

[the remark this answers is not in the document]
klauer: My first thought was a warning message like you said, but they are somewhat heavy/annoying to use (in my opinion) for things which might get called repeatedly like this.

`No bSetDefault call -> bValid:=FALSE` seems like it might be enough in this scenario, since it won't allow motion and should cover the majority of mistaken copy/paste errors.

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
```diff
@@ -0,0 +1,107 @@
...
+    fbStateSetup(stPositionState:=astStates1[3], sName:='TT', fPosition:=30);
+
+    fbStateSetup(stPositionState:=astStates2[1], sName:='OUT', fPosition:=-30);
+    fbStateSetup(stPositionState:=astStates2[2], sName:='YAG', fPosition:=35);
+    fbStateSetup(stPositionState:=astStates2[3], sName:='TT', fPosition:=70);
+
+    fbStateSetup(stPositionState:=astStates3[1], sName:='OUT', fPosition:=0.4, fDelta:=0.1);
+    fbStateSetup(stPositionState:=astStates3[2], sName:='YAG', fPosition:=2.3, fDelta:=0.1);
+    fbStateSetup(stPositionState:=astStates3[3], sName:='TT', fPosition:=5.6, fDelta:=0.1;
+*)
```
slactjohnson, line 48: Is it intended that this is only called during initialization/startup? I think that the example, as written, would continually over-write the values, preventing any modifications from EPICS. Is this something we want to prevent? 

Regardless, the code looks good. I like it.

slactjohnson (approved):
One question about intended use, but it LGTM regardless.

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
ZLLentz: You're right, the example as-written will implicitly block EPICS from making further changes. I think it makes sense to adjust the examples and make a note of why everything is wrapped inside a "call me only once" block.

ZLLentz: I guess one more follow-up thought is: is there any possible way for the function block itself to exercise a "do the thing only once" sort of behavior? It's worth a short brainstorm I think, since bInit only-run-me-once blocks are at least a bit annoying to write. Time to ponder...

ZLLentz: A simple adjustment to the function block code is to skip overwriting structs that already have writes to either the sName or bValid fields

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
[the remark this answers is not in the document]
ZLLentz: image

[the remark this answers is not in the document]
klauer: Well that looks awesome! Nicely done (with the `bHasWarned` stuff too)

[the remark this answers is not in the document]
ZLLentz: I tried to resolve this with https://github.com/pcdshub/lcls-twincat-motion/pull/184/commits/2d55fa2d20a6f75e5579276fa11a031197883ac4

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
ZLLentz: I tried to resolve this with https://github.com/pcdshub/lcls-twincat-motion/pull/184/commits/809f126477d7c7a9b5cdc90d93dcdb2303a95693

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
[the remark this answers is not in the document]
ZLLentz: I considered for a moment if I can include other info in the payload, but all I have custody of is the state name, I don't have access to the motor name for example. I'm not sure if it's worth including that info in the warning message.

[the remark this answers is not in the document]
ZLLentz: Revising my brain a bit: I can include the pmps string (if provided) which would help someone track down the FB.

lcls-twincat-motion/Library/POUs/Motion/States/FB_StateSetupHelper.TcPOU
klauer: I think that's a reasonable tweak, Zach.

I'm personally somewhat in favor of the `bInit` mechanism (at least partly because I can reset `bInit` and make it redo the initialization steps whenever I choose), but I understand the desire to use it as you mention @slactjohnson . This seems a reasonable middle-ground for both use cases.

slactjohnson: @klauer I don't really favor one way or the other; I see advantages to both. But in principle you could toggle `bForceUpdate`, right? And that way, you only re-init one thing.

klauer: `bForceUpdate` is per function block call rather than implicitly for all of them through an `IF bInit THEN` block. 

If you tied a global `bReinitialize` variable to all `bForceUpdate` inputs (or once as like `fbSetDefaults.bForceUpdate := bReinitialize;`) it'd even out in the end, I think

ZLLentz: I think the reason I went with this route instead of doing a docs revision is because calling these multiple times will break mirror-like state blocks that set their positions based on encoder counts. If there wasn't a potential PLC bug associated with running this twice I think the explicit bInit version is preferable.

The without this, the flow for the mirror-likes would be something like:

- set fPosition to 0
- set nEncoderCount to <some number>
- states FB calculates corrected fPosition value, and sets it
- various elements use the calculated fPosition
- next cycle: set fPosition to 0 (maybe even _during_ a move!)

So I'd like to avoid this kind of behavior altogether

ZLLentz: I think actually bForceUpdate runs on all of them for one FB since it doesn't get reset anywhere in the function block call

ZLLentz:
I'm going to merge this and move on to the next item

PR #192, opened 2023-08-24
API: remove many unused PVs that scale with motor/state counts

ZLLentz:
## Description

https://docs.google.com/spreadsheets/d/1xfSB1mJwVu4tXvxBT85TEYVi29Z7VoQ4w5CXw13rV8A/edit#gid=0

## Motivation and Context

> Due to issues with PV count and long boot times in the TMO plc motion IOC, I'm planning to remove a bunch of PVs from the twincat motion library. This should improve the boot time for all the other beckhoff motion IOCs too

## How Has This Been Tested?

N/a, hopefully the pragma lint CI passes

Google doc and release notes

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT
+++ b/lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT
@@ -9,40 +9,12 @@ STRUCT
     // PLC Axis Reference
     Axis: AXIS_REF;
     // NC Forward Limit Switch: TRUE if ok to move
-    {attribute 'pytmc' := '
-        pv: PLC:bLimitForwardEnable
-        io: i
-        field: ZNAM FALSE
-        field: ONAM TRUE
-        field: DESC FALSE if forward limit hit
-    '}
     bLimitForwardEnable AT %I*: BOOL;
     // NC Backward Limit Switch: TRUE if ok to move
-    {attribute 'pytmc' := '
-        pv: PLC:bLimitBackwardEnable
-        io: i
-        field: ZNAM FALSE
-        field: ONAM TRUE
-        field: DESC FALSE if reverse limit hit
-    '}
     bLimitBackwardEnable AT %I*: BOOL;
     // NO Home Switch: TRUE if at home
-    {attribute 'pytmc' := '
-        pv: PLC:bHome
-        io: i
-        field: ZNAM FALSE
-        field: ONAM TRUE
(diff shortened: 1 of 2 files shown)


klauer (approved):
I think this matches what's in the spreadsheet - and think what was culled there is reasonable.

Though as you've noted/requested, more opinions would be great to ensure we're not missing something.

ZLLentz:
I'm going to merge on Friday if no more thoughts roll in

ZLLentz:
One thing I noticed today that might feed back into this:
It seems like the limit indicators on the motor record are actually the "final" "anything has said I can't go forward/backward" indicators.
So, if a device has restrictions placed by EPS, gantry, etc., it all gets combined into those LLS HLS values.
So it might be good to bring back the specific enable PVs since they may get use in UIs later.

slactjohnson:
I don't have any issues with the removals. 

However, one thought I had is that if we're just trying to cut overall PV counts, maybe some of these boolean status PVs could be put together into a bit vector PV, kind of like the motor record MSTA field?

slactjohnson (approved):
No issues from me with the removals. LGTM.

ZLLentz:
You're 100% right Tyler, we could make a second MSTA and/or we could batch stuff like forward enables, reverse enables, etc. together

ZLLentz:
I think there some significant value in bitmask PVs specifically for these directional enables to help disambiguate why LLS/HLS are hit. If there is some agreement I can implement those pretty easily.

klauer:
I think it's a very reasonable addition 👍

ZLLentz:
I'm going to create an issue that proposes these bitmasks and suggests bundling it with the above work.
I'm going to merge this on Monday and make a "low PV count" tag, and work on the bitmask idea later.

ZLLentz:
@divyak1910 had a good point on the google doc RE: PVs that will be useful for configuration user patch panel motors, such as the homing mode (e.g. home to lls vs hls vs etc.)

I think what I want to do for these is to create a new ST_MotionStageUser that extends ST_MotionStage and adds some pytmc control points. I feel comfortable doing this because there's still a lot of other library changes needed to make user stages viable, including control points for NC and CoE parameters that don't belong on beamline motors.

To that end I think the removals here are all still correct- aside from bringing back some at bitmasks in a follow-up PR.

ZLLentz:
I am going to merge and tag this so we can set it loose on TMO motion and cut into that 22k PV count a little bit.

PR #198, opened 2023-09-07
FIX: addressing the state input deadlock bug

### LINKED ISSUE #197: Unknown Race Condition/State Machine Deadlock in States Motion
## Current Behavior

TMO's AL1K4 and LI1K4 ended up in a bad state deadlock where they could not move.

There are two function blocks I'd like to point out here.

First: https://github.com/pcdshub/lcls-twincat-motion/blob/master/lcls-twincat-motion/Library/POUs/Motion/States/FB_StatesInputHandler.TcPOU
Where:

- variable nSTATE = IDLE
- output bExecMove = TRUE
- input bMoveBusy = FALSE

Second:  https://github.com/pcdshub/lcls-twincat-motion/blob/master/lcls-twincat-motion/Library/POUs/Motion/FB_MotionRequest.TcPOU 
Where:

- variable nState = WAIT_EXEC
- input bExecute = TRUE (set from previous bExecMove output)
- output bDone = FALSE (indirectly used as the bMoveBusy input above)

The problems are:

- The motion state machine was waiting for an execute rising edge, and the input was giving out a constant execute TRUE that would never be able to rise.
- In this state, there is no escape hatch for the input handler to drop bExecMove back down to FALSE. As far as it's aware, we're still waiting on the previous move to end.
- In this state, there is no condition for the move state machine to say "I'm waiting for an exec, I currently have an exec, let's move on". As far as it is aware, the execute input is stale and it needs to wait for a new input.
- I don't know how we got into this state: it needs more investigation.

This is recoverable with an online write to bExecMove := FALSE

## Expected Behavior

States should always move when asked, if possible and safe.

## Context / environment

lcls-twincat-motion 4.0.1
lcls-twincat-common-components 3.0.1

## Steps to Reproduce (for bugs)

Unknown

## Suggested Solution

1. Understand how we get into this state deadlock and prevent it
2. Add an escape hatch to escape the deadlock- perhaps we can adjust some things when the FB errors are cleared?

ZLLentz:
## Description

When the user resets an error, bring the states input handler FB back to a known state.

## Motivation and Context

Will close #197 

In short: previously, we could get into a deadlocked state where the input FB would always keep bExecute high, which means the move FB would never be able to get a rising edge.

## How Has This Been Tested?

- Unit test that fails before and passes afterwards
- Interactive testing with the example motion plc code

Detailed documentation in the unit test

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/States/FB_StatesInputHandler.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/FB_StatesInputHandler.TcPOU
@@ -42,13 +42,15 @@ END_VAR
 ]]></Declaration>
     <Implementation>
       <ST><![CDATA[
-IF NOT bInit THEN
+bResetMove := stUserInput.bReset;
+IF bResetMove OR NOT bInit THEN
     bInit := TRUE;
     stUserInput.nSetValue := 0;
     nCurrGoal := nStartingState;
+    bExecMove := FALSE;
+    nState := IDLE;
+    bNewMove := FALSE;
 END_IF
-
-bResetMove := stUserInput.bReset;
 
 IF stUserInput.nSetValue <> 0 THEN
     nQueuedGoal := stUserInput.nSetValue;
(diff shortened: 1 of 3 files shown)


tylerjohnson3208 (approved):
LGTM.

tylerjohnson3208:
Whoops... logged in my personal account.

PR #201, opened 2023-09-20
FIX: run fbInternal first to mark states as ready to use

ZLLentz:
## Description

Calling read/move requires fb internal to be called first, but somehow the ordering was wrong

## Motivation and Context

Ends up making the pmps states devices come up in a "transition" state always because the PLC can't check that they are at the correct state during the first cycle

## How Has This Been Tested?

Currently testing using my kfe motion PR

n/a

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateND_Core.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateND_Core.TcPOU
@@ -51,6 +51,11 @@ END_VAR]]></Declaration>
       <ST><![CDATA[
 stEpicsToPlc.nSetValue := eEnumSet;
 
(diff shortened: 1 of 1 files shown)


tangkong (approved):
Hey I understood this one 👍

ZLLentz:
Running live, this fixes most of the issues but a few of my states devices still say no beam on startup and need a closer look

klauer (approved):
Makes sense that the initialization needs to happen first with `fbInternal` prior to `fbRead` and such handling it.

How do the unit tests fare? 

LGTM, even if it doesn't solve all issues 👍

ZLLentz:
It seems like there are further race conditions that need investigating separately
Fixing this means we're no longer guaranteed to have a bad startup but it happens sometimes anyways

PR #203, opened 2023-09-20
FIX: some fast faults could only be reset once due to rt detection

### LINKED ISSUE #202: FB_MotionBPTM does not properly allow for fault resetting
## Current Behavior

Sets the reset bits to TRUE under certain cases, but never drops them to FALSE in any cases
Since the fast fault block is looking for a rising edge reset, this means that these FBs can only be reset exactly once

## Expected Behavior

We should be able to reset these fast faults many times as needed

## Suggested Solution

Go through all the fast fault handling in this repo and make sure that we don't hold the reset flag high

ZLLentz:
## Description

A fast fault can be reset on a rising trigger to the reset input
Previously, the code would never bring this input back to low, it would stay high
Therefore, the fault could only be reset once

## Motivation and Context

Permanent fast faults are bad
closes #202 

## How Has This Been Tested?

Interactively in kfe motion, and:
image

n/a

## Pre-merge checklist


klauer (approved):
Logic seems sound to me 👍

PR #205, opened 2023-10-06
FIX: issues relating to "queued" moves in state mover with bMoveOk

ZLLentz:
## Description

- If there is a move error, cut the execute output
- Require a bExecute rising edge and bAllowMoves at the same time, or else no state moves

## Motivation and Context

See https://jira.slac.stanford.edu/browse/ECS-3932
There are cases where "unsafe" moves can become marked as "safe" during execution, and then immediately start as if they were waiting to become "safe" rather than cancelling the move request altogether.

## How Has This Been Tested?

I created a unit test that follows the following sequence:
- Ask for a move to a `bMoveOk = FALSE` state
- Confirm no move
- Set `bMoveOk := TRUE`
- Confirm no move
- Reset the error
- Confirm no move
- Ask for a new move
- Confirm yes move

I also tested this interactively, and it works as one would expect: you need to reset the error, wait for the move to become safe, and then explicitly ask for the move again. Here's a screenshot of the new updated error message:
image

Release notes only

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
@@ -139,7 +139,7 @@ IF nLatchAllowErrorID <> 0 OR (bExecute AND NOT bAllowMove) THEN
     END_IF
     // Keep error latched until it is cleared, otherwise it can be lost early
     nLatchAllowErrorID := nErrorID;
-    sErrorMessage := CONCAT(CONCAT(F_MotionErrorCodeLookup(nErrorId := nErrorID), " for "), stPositionState.sName);
+    sErrorMessage := CONCAT(CONCAT(F_MotionErrorCodeLookup(nErrorId := nErrorID), ' for '), stPositionState.sName);
 END_IF
 
 // This can be useful if we're running this FB standalone for some reason
--- a/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
@@ -91,17 +91,25 @@ VAR_OUTPUT
 END_VAR
 VAR
     fbMotionRequest: FB_MotionRequest;
+    rtExec: R_TRIG;
+    rtReset: R_TRIG;
+    bInnerExec: BOOL;
     bAllowMove: BOOL;
+    nLatchAllowErrorID: UDINT;
 END_VAR
 ]]></Declaration>
     <Implementation>
       <ST><![CDATA[// Veto the move for uninitialized and unsafe states
 bAllowMove := stPositionState.bMoveOk AND stPositionState.bValid AND stPositionState.bUpdated;
 
+rtExec(CLK:=bExecute);
+bInnerExec S= rtExec.Q AND bAllowMove;
+bInnerExec R= NOT bExecute;
+
 // Do the move
 fbMotionRequest(
     stMotionStage := stMotionStage,
-    bExecute := bExecute AND bAllowMove,
+    bExecute := bInnerExec,
     bReset := bReset,
     enumMotionRequest := enumMotionRequest,
     fPos := stPositionState.fPosition,
@@ -114,15 +122,24 @@ fbMotionRequest(
     bBusy => bBusy,
     bDone => bDone);
 
+rtReset(CLK:=bReset);
+IF rtReset.Q THEN
+    nLatchAllowErrorID := 0;
+END_IF
+
 // Inject custom error if we can't move because of bMoveOk or bValid
-IF bExecute AND NOT bAllowMove THEN
+IF nLatchAllowErrorID <> 0 OR (bExecute AND NOT bAllowMove) THEN
     bError := TRUE;
-    IF stPositionState.bValid THEN
-        nErrorId := 16#7901;
+    IF nLatchAllowErrorID <> 0 THEN
+        nErrorID := nLatchAllowErrorID;
+    ELSIF stPositionState.bValid THEN
+        nErrorID := 16#7901;
     ELSE
-        nErrorId := 16#7902;
+        nErrorID := 16#7902;
     END_IF
-    sErrorMessage := F_MotionErrorCodeLookup(nErrorId := nErrorID);
+    // Keep error latched until it is cleared, otherwise it can be lost early
+    nLatchAllowErrorID := nErrorID;
+    sErrorMessage := CONCAT(CONCAT(F_MotionErrorCodeLookup(nErrorId := nErrorID), " for "), stPositionState.sName);
 END_IF
 
 // This can be useful if we're running this FB standalone for some reason
--- a/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
@@ -133,9 +133,9 @@ IF nLatchAllowErrorID <> 0 OR (bExecute AND NOT bAllowMove) THEN
     IF nLatchAllowErrorID <> 0 THEN
         nErrorID := nLatchAllowErrorID;
     ELSIF stPositionState.bValid THEN
-        nErrorID := 16#7901;
+        nErrorID := E_LCLSMotionError.UNSAFE;
     ELSE
-        nErrorID := 16#7902;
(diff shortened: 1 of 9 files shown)


ZLLentz:
I'm a bit stumped as to the failure state of my test- at the end, the motor moves as I expect it to, but the function block reports that it is not busy. I'm not sure what I'm looking at, if it's a bug related to this code or unrelated or not a bug at all and I'm calling the code wrong.

ZLLentz:
It turns out that my tests were failing simply because I wasn't waiting long enough to check the results. I feel a little bit silly.

ZLLentz:
The tests pass without the patch, meaning that I've completely failed at some part of the dev process here.
I need to go back to interactive pre-patch testing to make sure I understood the issue properly.

ZLLentz:
I'm currently taking this as a learning experience to experiment with and get better at writing these sorts of unit tests. Hopefully this is a time investment that will pay off later. I have not entirely figured out what is wrong, but it has something to do with extra moves being requested on my motor that I don't intend to make.

ZLLentz:
This newest version of the test consistently passes without hitting the weird errors I was seeing, largely by being less "cycle-precise" than the previous iterations. That is, we're hitting buttons more like "humans" with human delays.

I think this means that some of these function blocks have race conditions, and I think some of these race conditions arise from the interaction of the NC process with the PLC process.

In addition, the test now fails how I expected it to fail with the pre-patch code.

ZLLentz:
"The usual suspects" for code review are all at a conference right now- feel free to wait until next week to take a look.

ZLLentz:
@klauer @slactjohnson @ghalym I'd appreciate at least one review on this so we can close https://jira.slac.stanford.edu/browse/ECS-3932

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
```diff
@@ -114,15 +122,24 @@ fbMotionRequest(
...
 // Inject custom error if we can't move because of bMoveOk or bValid
-IF bExecute AND NOT bAllowMove THEN
+IF nLatchAllowErrorID <> 0 OR (bExecute AND NOT bAllowMove) THEN
     bError := TRUE;
-    IF stPositionState.bValid THEN
-        nErrorId := 16#7901;
+    IF nLatchAllowErrorID <> 0 THEN
+        nErrorID := nLatchAllowErrorID;
+    ELSIF stPositionState.bValid THEN
+        nErrorID := 16#7901;
```
klauer, line 136: A comment on these magic numbers would be appreciated (or eventually moved to an enum as we *may* have previously chatted about, I think)

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
```diff
@@ -114,15 +122,24 @@ fbMotionRequest(
...
     bDone => bDone);
 
+rtReset(CLK:=bReset);
+IF rtReset.Q THEN
+    nLatchAllowErrorID := 0;
+END_IF
+
 // Inject custom error if we can't move because of bMoveOk or bValid
-IF bExecute AND NOT bAllowMove THEN
+IF nLatchAllowErrorID <> 0 OR (bExecute AND NOT bAllowMove) THEN
```
klauer, line 131: In the scenario of `nLatchAllowErrorID <> 0`, it's necessary to redo this `IF` block not to re-look things up, but ensure that `bError` is set (no matter what `fbMotionRequest` does to it), is that right? Trying to follow the logic, and it seems largely a no-op beside for that.

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
```diff
@@ -91,17 +91,25 @@ VAR_OUTPUT
...
+    nLatchAllowErrorID: UDINT;
 END_VAR
 ]]></Declaration>
     <Implementation>
       <ST><![CDATA[// Veto the move for uninitialized and unsafe states
 bAllowMove := stPositionState.bMoveOk AND stPositionState.bValid AND stPositionState.bUpdated;
 
+rtExec(CLK:=bExecute);
+bInnerExec S= rtExec.Q AND bAllowMove;
+bInnerExec R= NOT bExecute;
```
klauer, line 107: I'd naively have expected if there were a latched error we wouldn't execute `fbMotionRequest` at all. I'd imagine there are some details I'm missing there - mind giving a quick explanation?

lcls-twincat-motion/Library/POUs/Motion/States/FB_StatesInputHandler.TcPOU
```diff
@@ -84,6 +86,10 @@ CASE nState OF
         END_IF
 END_CASE
 
+IF bMoveError THEN
+    bExecMove := FALSE;
```
klauer, line 90: Ah, I think this is the line that explains my above question...

klauer (dismissed):
The latching logic makes sense to me, I think - though I do want to confirm that with the few comments I have here.

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
ZLLentz: I thought I'd done the background work for this and forgot to apply it here, but the reality is that I only implemented enums so far for the FF types. The custom error codes are reflected only the error message writer FB, this needs some better organization as you've said before.
https://github.com/pcdshub/lcls-twincat-motion/blob/master/lcls-twincat-motion/Library/DUTs/E_MotionFFType.TcDUT
https://github.com/pcdshub/lcls-twincat-motion/blob/master/lcls-twincat-motion/Library/POUs/Motion/Utils/F_MotionErrorCodeLookup.TcPOU

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
ZLLentz: That's exactly correct. I need to reapply my own FB's error status, including `bError` and `nErrorID`, irrespective of the internal error state of `fbMotionRequest`.

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
ZLLentz: I think, despite your other comment, you're correct. It may be worth doing something additional here, making the inner function blocks also protect themselves for the sake of robustness.

lcls-twincat-motion/Library/POUs/Motion/States/FB_StatesInputHandler.TcPOU
ZLLentz: This is indeed why it works out, but I think your previous point still stands

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
ZLLentz: My intuition is that this calls for a new enum specifically for custom error types- not to rehash all the built-in errors.

klauer: I think the "custom error code enum" makes a lot of sense. This certainly could be in a follow-up PR (if at all)

ZLLentz: Handled in https://github.com/pcdshub/lcls-twincat-motion/pull/205/commits/1170c26791cf8bf6f5f0c0f8a2e1011ab3e221c7

lcls-twincat-motion/Library/POUs/Motion/States/FB_PositionStateMove.TcPOU
ZLLentz: Handled in https://github.com/pcdshub/lcls-twincat-motion/pull/205/commits/cc683b2eff1adc02ef9082efaa88601d23c5522e

klauer (approved):
The enum additions look good and I have no further suggestions / nits

I'd say this is good to go, though it'd be nice to have another set of eyes on it for good measure

ZLLentz:
I'm fairly confident in this and I'm already here, so I'm going to merge, tag, and deploy the tag. If something else is discovered between now and the PAMM there is time to address it.

PR #209, opened 2024-02-02
BUG: reverse the in/out order in the in/out example fbs to match their enum

### LINKED ISSUE #208: IN and OUT reversed in provided in/out FB
## Current Behavior

The enum has OUT at state 1 and IN at state 2
FB_PositionStage1D_InOut puts stIn into state 1 and stOut into state 2
This is backwards

## Expected Behavior

Keep the enum the same
Reverse the array filling in the example fill

## Context / environment

Causes AT1K2 axis 3 to have confusing behavior (IN and OUT are reversed)

ZLLentz:
## Description

Reverse IN and OUT indices for the example state FBs to match the enum, because if the order is wrong then IN sends us to OUT and vice versa.
See https://github.com/pcdshub/lcls-twincat-motion/blob/master/lcls-twincat-motion/Library/DUTs/E_EpicsInOut.TcDUT

## Motivation and Context

closes #208 

## How Has This Been Tested?

Fixes the remaining issue in https://github.com/pcdshub/lcls-plc-kfe-rix-motion/pull/61

only here

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/States/Examples/FB_PositionState1D_InOut.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/States/Examples/FB_PositionState1D_InOut.TcPOU
@@ -61,8 +61,8 @@ fbPositionState1D(
 );
 
 // Send updates made on the array back to the inputs (VAR_IN_OUT) for maximum clarity
-stIn := astPositionState[1];
-stOut := astPositionState[2];]]></ST>
+stOut := astPositionState[1];
+stIn := astPositionState[2];]]></ST>
     </Implementation>
   </POU>
 </TcPlcObject>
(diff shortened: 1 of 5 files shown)


lcls-twincat-motion/Library/POUs/Motion/States/Examples/FB_PositionState1D_InOut.TcPOU
```diff
@@ -47,9 +47,9 @@ IF stOut.sName = '' THEN
     stOut.sName := 'OUT';
 END_IF
 
-// Assemble the states array, matching the enum values (IN is 1, OUT is 2)
-astPositionState[1] := stIn;
-astPositionState[2] := stOut;
+// Assemble the states array, matching the enum values (OUT is 1, IN is 2)
+astPositionState[1] := stOut;
```
ZLLentz, line 51: self-review thought before I leave today: this should probably look more like this:
```
astPositionState[E_EpicsInOut.OUT] := stOut
```

ZLLentz:
The unit tests pass, provided you don't run them on the BSD PLCs... this is new to me and something I'll have to resolve as part of the TwinCAT CI project.

I'd appreciate if I could get 1 sanity check on this. The diff is pretty straightforward and doesn't require a full understanding of the library.

PR #214, opened 2024-03-19
ENH: invesitgate flutter fix

nrwslac:
## Description

- Implemented a function block  that can detect variables that flip too much, causing a flutter. Use that function block to show how it could be used to take away auto-reset capabilities for a particular fast fault. I am not sure this is a good solution, but worth discussing; seems like a lot of overhead. Probably a better solution is updating the GUI to show fluttering fast faults as discussed with @ZLLentz. 

- Also, unsure where we would want to trigger a reset of `fbFlutterDet`. I guess we want to give back the auto-reset capability when the flutter condition is gone. 
 
## Motivation and Context

-Its hard to bypass a fluttering fast fault.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/Library.plcproj
+++ b/lcls-twincat-motion/Library/Library.plcproj
@@ -597,9 +597,9 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
-    <PlaceholderReference Include="LCLS General">
-      <DefaultResolution>LCLS General, * (SLAC)</DefaultResolution>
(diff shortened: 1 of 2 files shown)


ZLLentz:
I think we need to do both: defluttering where we know about it, and update the gui as well so we can work around new cases that we don't know about yet

nrwslac:
> I think we need to do both: defluttering where we know about it, and update the gui as well so we can work around new cases that we don't know about yet

I'll try to look at the filtering next.

PR #215, opened 2024-03-22
ENH: check PLC OS to set new default dir for BSD PLCs

ZLLentz:
## Description

Use the IPCDiag FBs to check the PLC's OS, then use that OS to decide the default directory for pmpsdb reading.
These function blocks are somewhat strange, but I followed the docs closely and arrived at this implementation.

Here, I've set the new default directory for TwinCAT BSD PLCs to match the associated PRs:
https://github.com/pcdshub/twincat-bsd-ansible/pull/19
https://github.com/pcdshub/pmpsdb_client/pull/25

And I've left the default directory for other PLC types unchanged.

## Motivation and Context

We can explicitly set a directory for the PMPS database loading in each project, but this leaves open the possibility for typos and copy/paste errors. The directory should be the same every time (dependent on OS), so we can directly check the OS to set a reasonable default. The code is skipped entirely if the user provides a directory.

## How Has This Been Tested?

Interactively on a BSD PLC: the correct directory was checked and the file was opened. (plc-tst-bsd2)
Also checked on a Windows CE PLC, test suite passed and default directory was ftp (plc-tst-proto8, plc-tst-pmps-subsytem-b)

Tested again after adding some error handling, seemed to do the right thing

https://confluence.slac.stanford.edu/display/PCDS/TcBSD+Configuration+Reference#TcBSDConfigurationReference-PMPSDatabase

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
@@ -45,7 +45,8 @@ VAR
 
     fbIPCReg: FB_IPCDiag_Register;
     fbCheckOS: FB_IPCDiag_ReadParameter;
-    sOSName: STRING;
+    sOSName: STRING := '';
+    nCheckOSTries: UINT := 3;
 END_VAR
 ]]></Declaration>
     <Implementation>
@@ -68,13 +69,27 @@ END_IF
 
 IF sDirectory = '' THEN
     // Check OS for default directory
-    fbIPCReg(bExecute:=TRUE);
+    fbIPCReg(
+        bExecute:=TRUE,
+        tTimeout:=T#1s,
+    );
     fbCheckOS(
-        bExecute:=NOT fbIPCReg.bBusy,
+        bExecute:=fbIPCReg.bValid,
         eParameterKey:=E_IPCDiag_ParameterKey.OS_Name,
         fbRegister:=fbIPCReg,
     );
-    IF fbCheckOS.bValid THEN
+    IF fbIPCReg.bError OR fbCheckOS.bError THEN
+        IF nCheckOSTries <= 0 THEN
+            // Retry counter down to zero: set to error name to try old default
+            sOSName := 'Error';
+        ELSE
+            // Decrement retry counter
+            nCheckOSTries := nCheckOSTries - 1;
+            // Reset FBs
+            fbCheckOS(bExecute:=FALSE, fbRegister:=fbIPCReg);
+            fbIPCReg(bExecute:=FALSE);
+        END_IF
+    ELSIF fbCheckOS.bValid THEN
         fbCheckOS.GetParameter(
             pBuffer:=ADR(sOSName),
             nBufferSize:=SIZEOF(sOSName),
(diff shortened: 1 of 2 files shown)


ZLLentz:
I will run the unit tests right now, but I won't test the Windows CE PLC until Monday. When I'm in the office it will be more straightforward to avoid conflicts.

If it works on Windows CE then I'll be confident to mark as ready for review and ask for reviewers next week.

ZLLentz:
I'm having trouble accessing the prog node (my account is stuck in logging out?) but I'm not going to debug it after 4 on a Friday. I'll resume this next week.

ZLLentz:
This doesn't brick the test suite, but also I didn't expand the test suite here because I wasn't sure if there was anything meaningful to do. The only test I can think of is a sort of "reflexive" test where I duplicate the logic in the test suite and make sure its being followed but this doesn't have very much value in my opinion.

ZLLentz:
I think this is done, looking for opinions on the direction + code review + suggestions on some sort of useful unit test I can add

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
```diff
@@ -62,9 +66,30 @@ IF rtEnable.Q OR rtRefresh.Q THEN
...
 
+IF sDirectory = '' THEN
+    // Check OS for default directory
+    fbIPCReg(bExecute:=TRUE);
+    fbCheckOS(
+        bExecute:=NOT fbIPCReg.bBusy,
+        eParameterKey:=E_IPCDiag_ParameterKey.OS_Name,
+        fbRegister:=fbIPCReg,
+    );
+    IF fbCheckOS.bValid THEN
```
nrwslac, line 77: I noticed this includes a check for error, but if there is an error, we won't be flagged.

ZLLentz: Good catch. You're right, as-written if we have an error the JSON file will silently never load and we might not notice an issue until we check the PMPS diagnostic.

This warrants some revision.

ZLLentz: Some options that I'm tossing around right now:

- If there's an error, try again? (a few times?)
- Go back to the old default?
- Give an error log message?
- Fast fault?

Needs some thought

ZLLentz: I'm going to opt for 3 tries followed by going back to the old default
This gives us some wiggle room for first-cycle-bs and will give us some normal error messages if the old default was wrong and if the old default was correct then the PLC will load values as normal.

ZLLentz:
I'm going to click update branch and re-request reviews here, I forgot about this PR until I checked my jira tickets.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_Standard_PMPSDB.TcPOU
nrwslac: I think this means, if the OS isn't solved and a twincat PLC gets the default directory, fbPmpsFileReader will throw a directory not found error.

nrwslac (approved):
lgtm.

ZLLentz:
I'm going to merge and release this if there are no objections from @ghalym and @NSLentz prior to midday tomorrow

NSLentz (approved):
lgtm

PR #216, opened 2024-04-05
Implemented Changes to Expose Axis Parameters as READ-ONLY to EPICS

NSLentz:
## Description

Created a new DUT called ST_AxisParameterSetExposed.
Added the new DUT as a member of ST_MotionStage.
Added logic to FB_MotionStage to copy the exposed parameters into the new DUT whenever the stAxisParameters structure is updated.
Added pytmc pragmas to the ST_AxisParameterSetExposed to populate the new PVs when an IOC is built using lcls-twincat-motion.

## Motivation and Context

This new feature came up as a result of difficulties with scientists and other staff in understanding why certain motions cause errors or are not allowed to be completed. With these parameters exposed, they will have greater transparency with understanding the limits of the system that have been programmed into the PLC.

## How Has This Been Tested?

A new unit test was added to test that the values are being copied properly.
All the previous unit tests and the new unit test were run and they all pass.
The library was temporarily installed as version 0.0.0.
The 0.0.0 library version was selected in a test PLC project, and used to define a motion stage.
A test ioc was built on tst-console and a test screen was created to query the 10 PVs that were added.
Each PV was changed within the NC configuration and the test screen was watched to make sure the value updated to the correct new value within a reasonable amount of time (about 1 second).

The code has been commented thoroughly to explain the intended function of each added PV. In addition, the changes to FB_MotionStage have a code comment to describe the intended function of the new copying of parameters.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/DUTs/ST_AxisParameterSetExposed.TcDUT
+++ b/lcls-twincat-motion/Library/DUTs/ST_AxisParameterSetExposed.TcDUT
@@ -1,0 +1,76 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<TcPlcObject Version="1.1.0.1">
+  <DUT Name="ST_AxisParameterSetExposed" Id="{ee296551-543a-4bbb-98b7-123ff2229439}">
+    <Declaration><![CDATA[TYPE ST_AxisParameterSetExposed :
+    // Collects the axis parameters that are intentionally exposed to EPICS.
+STRUCT
+    // Maximum Dynamics
+    {attribute 'pytmc' := '
+        pv: MaxVel
+        io: i
+        field: DESC Maximum commandable speed of the axis in EU/s.
+    '}
+    fVeloMaximum                : LREAL; // Maximum commandable speed of the axis in EU/s.
+    {attribute 'pytmc' := '
+        pv: MaxAccel
+        io: i
+        field: DESC Maximum rate of increase in speed of the axis in EU/s^2.
+    '}
+    fAccelerationMax            : LREAL; // Maximum rate of increase in speed of the axis in EU/s^2.
+    {attribute 'pytmc' := '
+        pv: MaxDecel
+        io: i
+        field: DESC Maximum rate of decrease in speed of the axis in EU/s^2.
+    '}
+    fDecelerationMax            : LREAL; // Maximum rate of decrease in speed of the axis in EU/s^2.
+
+    // Monitoring
+    {attribute 'pytmc' := '
+        pv: PosLagEn
+        io: i
+        field: DESC TRUE if position lag monitor (also known as stall monitor) is enabled.
+    '}
+    bCtrlEnablePosDiffControl   : WORD;  // Enable/Disable state of Position Lag Monitor.
+    {attribute 'pytmc' := '
+        pv: PosLagVal
+        io: i
+        field: DESC Maximum magnitude of position lag in EU.
+    '}
+    fCtrlPosDiffMax             : LREAL; // Maximum magnitude of position lag in EU.
+    {attribute 'pytmc' := '
+        pv: PosLagTime
+        io: i
+        field: DESC Maximum allowable duration outside of maximum position lag value in seconds.
+    '}
+    fCtrlPosDiffMaxTime         : LREAL; // Maximum allowable duration outside of maximum position lag value in seconds.
+
+    // Limit Switches
+    {attribute 'pytmc' := '
+        pv: SLimMinEn
+        io: i
+        field: DESC TRUE if controller static minimum limit is enabled.
+    '}
+    bEncEnableSoftEndMinControl : WORD;  // Enable/Disable state of controller static minimum limit.
+    {attribute 'pytmc' := '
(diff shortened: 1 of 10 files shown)


lcls-twincat-motion/Library/DUTs/ST_AxisParameterSetExposed.TcDUT
```diff
@@ -0,0 +1,76 @@
...
+    <Declaration><![CDATA[TYPE ST_AxisParameterSetExposed :
+    // Collects the axis parameters that are intentionally exposed to EPICS.
+STRUCT
+    // Maximum Dynamics
+    {attribute 'pytmc' := '
+        pv: MaxVel
+        io: i
+        field: DESC Maximum commandable speed of the axis in EU/s.
+    '}
+    fVeloMaximum                : LREAL; // Maximum commandable speed of the axis in EU/s.
```
ZLLentz, line 13: I thought these needed to be above the parameter but I was wrong, these associate properly in the IDE as-is

lcls-twincat-motion/_Config/NC/Axes/Axis_AxisParameter_Test.xti
ZLLentz, line 1: I think this test is pretty good, catches all the cases for now
It feels a bit reflexive as we discussed yesterday but it will catch errors if someone edits the FB later and mixes up some existing parameters

lcls-twincat-motion/Library/DUTs/ST_AxisParameterSetExposed.TcDUT
```diff
@@ -0,0 +1,76 @@
...
+    '}
+    bCtrlEnablePosDiffControl   : WORD;  // Enable/Disable state of Position Lag Monitor.
+    {attribute 'pytmc' := '
+        pv: PosLagVal
+        io: i
+        field: DESC Maximum magnitude of position lag in EU.
+    '}
+    fCtrlPosDiffMax             : LREAL; // Maximum magnitude of position lag in EU.
+    {attribute 'pytmc' := '
+        pv: PosLagTime
```
ZLLentz, line 41: This is the longest new PV name and it's still shorter than the longest existing PV so we're good on PV lengths here :+1:
`:PLC:AxisPara:PosLagTime`
vs
`:PLC:bGantryBackwardEnable`

ZLLentz: Note for future: here's how I found the longest as-deployed PV name, with a big assist from stackoverflow for use of awk:
```
(pcds-5.8.2)zlentz@psbuild-rhel7-01:~$ grep_pv :PLC: | grep MMS | awk ' { if ( length > x ) { x = length; y = $0 } }END{ print y }'
SP1L0:KMONO:MMS:DIODE_HORIZ:PLC:bGantryBackwardEnable_RBV, "bi"
```

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStageNCParams.TcPOU
```diff
@@ -32,6 +32,19 @@ mcReadParams(
...
+stMotionStage.stAxisParametersExposed.bEncEnableSoftEndMaxControl   := stMotionStage.stAxisParameters.bEncEnableSoftEndMaxControl;
+stMotionStage.stAxisParametersExposed.bEncEnableSoftEndMinControl   := stMotionStage.stAxisParameters.bEncEnableSoftEndMinControl;
+stMotionStage.stAxisParametersExposed.fAccelerationMax              := stMotionStage.stAxisParameters.fAccelerationMax;
+stMotionStage.stAxisParametersExposed.fCtrlPosDiffMax               := stMotionStage.stAxisParameters.fCtrlPosDiffMax;
+stMotionStage.stAxisParametersExposed.fCtrlPosDiffMaxTime           := stMotionStage.stAxisParameters.fCtrlPosDiffMaxTime;
+stMotionStage.stAxisParametersExposed.fDecelerationMax              := stMotionStage.stAxisParameters.fDecelerationMax;
+stMotionStage.stAxisParametersExposed.fEncSoftEndMax                := stMotionStage.stAxisParameters.fEncSoftEndMax;
+stMotionStage.stAxisParametersExposed.fEncSoftEndMin                := stMotionStage.stAxisParameters.fEncSoftEndMin;
+stMotionStage.stAxisParametersExposed.fVeloMaximum                  := stMotionStage.stAxisParameters.fVeloMaximum;
```
ZLLentz, line 46: I think this is an appropriate selection
Maybe in the future people would be interested in more but these cover the essentials as relating to EPS

ZLLentz (approved):
This looks good to me- I tried to find something to change but I think it's good as-is. I left comments about the things I checked.

nrwslac:
I need to check, I don't think it exists, and if it doesn't already, can we add encoder scaling and offset? @NSLentz

NSLentz:
Added encoder scaling and offset to NC configuration. Requesting re-review to merge. Testing steps were repeated with an updated test case for the added parameters.

ZLLentz (approved):

PR #217, opened 2024-05-14
Fixing Bug ECS 4719 Pertaining to PMPS Position State Limits Being Invalidated when State Switches to Unknown

NSLentz:
## Description

Added bLockBounds boolean to FB_StatePMPSEnables to lock the position limits active of the last valid goal state. Even if the axis moves out of the state without using the state mover, the position limits will remain active until a new valid goal state is selected. The bLockBounds boolean is kept true after the first valid goal state until the bEnable bit for the function block set to false. At that time, the bLockBounds boolean is reset.
Made it so that the bPowerSelf bit gets automatically set to FALSE if the FB_StatePMPSEnables block is used. This is irrespective of whether the bEnable bit for the FB_StatePMPSEnables block is true or false. If a FB_StatePMPSEnables block is called in a project the assumption is that bPowerSelf must be set to false. The alternative requires users to remember to set bPowerSelf to false and in practice this gets missed a lot.
Made it so that the PMPS position limits also prevent commands to positions outside of the allowable range instead of only looking at the current position. This will make it more difficult to command the axis out of the state fDelta range. Added formatted string error message to make it more clear when the PMPS position state limits are preventing a move.
Added a new error id to the F_MotionErrorCodeLookup. This is the error id for invalid I/O data for more than n continuous NC cycles (encoder).

## Motivation and Context

Fixed Bug https://jira.slac.stanford.edu/browse/ECS-4719 (ECS-4719) from Jira.

## How Has This Been Tested?

Tested by running full list of test cases. Added two new test cases without the bug fix incorporated. Verified the added test cases replicated the bug. Then fixed the bug and verified the test cases now passed. Needed to modify set and move helper function to make sure it initializes the fPosition value to the goal position if the axis is already at the goal position without trying to move. Added a new set and move helper function that is simpler to understand and parse.

Documented in various code comments and also made a small update to the README to document that bPowerSelf is automatically set false when using the PMPS enables function block.

## Pre-merge checklist


ZLLentz:
I think this changes enough of a central component that I also want to try it myself before hitting approve, but I'll start by reviewing the code line-by-line

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_StatePMPSEnables.TcPOU
```diff
@@ -129,9 +139,17 @@ ffNoGoal(
...
     the current position and the calculated bounds.
 *)
-IF bValidGoal AND bEnable THEN
-    bForwardEnabled := stMotionStage.stAxisStatus.fActPosition < fUpperPos;
-    bBackwardEnabled := stMotionStage.stAxisStatus.fActPosition > fLowerPos;
+IF bLockBounds THEN
+    // Prevent forward/backward motion if the position of the axis is outside of the upper/lower bounds respectively.
+    // Prevent forward/backward motion if the command to the axis is outside of the upper/lower bounds respectively.
+    bForwardEnabled := stMotionStage.stAxisStatus.fActPosition < fUpperPos AND stMotionStage.fPosition < fUpperPos;
+    bBackwardEnabled := stMotionStage.stAxisStatus.fActPosition > fLowerPos AND stMotionStage.fPosition > fLowerPos;
```
ZLLentz, line 146: I'm waffling a bit in my head if it might be better to use the direct sources for these numbers rather than the EPICS status and setpoint fields. In practice these are essentially always the same but I have to wonder if there are any edge cases where either this or the raw values as reported by the NC are more correct to use.

Maybe this is best as-is due to interface flexibility (technically not requiring the NC)

But the other option would have been to use `stMotionStage.Axis.NcToPlc.ActPos` and `stMotionStage.Axis.NcToPlc.TargetPos`

Food for thought, almost didn't even post this

Edit: another thing to consider is that fPosition here is not valid until bExecute goes to TRUE, which feels like a rare edge case that could occur if someone changed code elsewhere.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_StatePMPSEnables.TcPOU
```diff
@@ -129,9 +139,17 @@ ffNoGoal(
...
+
+    IF stMotionStage.nErrorId = 16#4223 AND (NOT bForwardEnabled OR NOT bBackwardEnabled) THEN
+        // Cannot move forward or backward, overwrite the error with a custom error message to give more context.
+        stMotionStage.sCustomErrorMessage := CONCAT(CONCAT(CONCAT(
+            'Limits exceeded for most recent PMPS position state: ', LREAL_TO_FMTSTR(fLowerPos,5,TRUE)), ' < pos < '), LREAL_TO_FMTSTR(fUpperPos,5,TRUE));
```
ZLLentz, line 151: It's a somewhat absurd edge case, but this error isn't necessarily correct if the user wants to move positive and the PMPS restriction is "bForwardEnable = True, bBackwardEnabled = False" but instead they are hitting the positive limit to bring forth the 16#4223 error, or otherwise the general enable is False.

I'd suggest matching the directional enables with the last setpoint vs the current position.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_StatePMPSEnables.TcPOU
```diff
@@ -98,13 +103,18 @@ mc_power(
 IF nGoalStateIndex > 0 AND nGoalStateIndex <= GeneralConstants.MAX_STATES THEN
     IF astPositionState[nGoalStateIndex].bValid AND astPositionState[nGoalStateIndex].bUpdated THEN
         bValidGoal := TRUE;
+        bLockBounds := TRUE;
```
ZLLentz, line 106: This means that we stay locked to the most recent setpoint at all times, right? I think it's a good implementation, solid and not overcomplicated.

lcls-twincat-motion/Library/Tests/FB_StatePMPSEnables_Test.TcPOU
```diff
@@ -31,8 +31,7 @@ VAR
 END_VAR
 ]]></Declaration>
     <Implementation>
-      <ST><![CDATA[
-astPositionState[nGoalState].fPosition := 10;
+      <ST><![CDATA[astPositionState[nGoalState].fPosition := 10;
```
ZLLentz, line 34: The start of file and end of file newlines are intentional to make the code more readable in review here, you don't have to include them but you probably shouldn't be removing them.

lcls-twincat-motion/Library/Tests/FB_StatePMPSEnables_Test.TcPOU
```diff
@@ -288,6 +292,136 @@ IF tonTimer.Q OR fbMove.bSetDone THEN
...
+
+bInit := TRUE;
+
+fbFFHWO.EvaluateOutput();
+
+// If we've reached the position OR ran out of time to set the position, check the asserts
+IF tonTimer.Q OR fbMotionStageSetAndMove.bMoveDone THEN
+    AssertTrue(
+        tonTimer.Q,
+        'Timeout should have occurred. Move should have been unable to complete.',
```
ZLLentz, line 370:
```suggestion
        'Timeout should not have occurred. Move should have been unable to complete.',
```

lcls-twincat-motion/_Config/NC/NC.xti
```diff
@@ -34,6 +34,7 @@
 		<Axis File="Axis_PositionStatePMPSND_Test_1.xti" Id="17"/>
 		<Axis File="Axis_PositionStatePMPSND_Test_2.xti" Id="18"/>
 		<Axis File="Axis_PositionStatePMPSND_Test_3.xti" Id="19"/>
+		<Axis File="Axis_StatePMPSEnable_Test_1.xti" Id="23"/>
 		<Axis File="Axis_StatePMPSEnable_Test.xti" Id="20"/>
```
ZLLentz, line 38: What's with the second motor here? Is one of them unused?

ZLLentz (commented):
I think this is good, I read through the code and the test carefully and pointed out the things that I had comments on.
I'm going to run this on my test plc and see how it feels as a user.

ZLLentz:
Snazzy so far, will check a few more things in a bit
image

ZLLentz:
Worked well in practice, can merge if it still passes tests after update branch and with comments addressed

lcls-twincat-motion/Library/Tests/FB_StatePMPSEnables_Test.TcPOU
NSLentz: The test case intentionally checks to make sure that it did timeout. E.g., the only reason it wouldn't time out is if it completed the move but completing the move is actually the failure condition. So the timeout should occur.

ZLLentz: Whoops, I realized this about halfway through reviewing the test but left this comment in anyway

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_StatePMPSEnables.TcPOU
NSLentz: I made it so that this special error message will only display if the bAllForwardEnable/bAllBackwardEnable from stMotionStage are true and they line up with the correct PMPS direction enable

lcls-twincat-motion/Library/Tests/FB_StatePMPSEnables_Test.TcPOU
NSLentz: Oh okay I didn't know that was the reasoning. I will fix.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_StatePMPSEnables.TcPOU
NSLentz: I switched it to use the NC values directly. I think that makes sense

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_StatePMPSEnables.TcPOU
ZLLentz: Looks pretty comprehensive to me :+1:

ZLLentz (dismissed):
I think all that's left is to click the "update branch" button that's blocking the merge button and re-run the unit tests while merged with all other changes from master

ZLLentz (approved):

ZLLentz:
Admin merge to bypass CI issue after Nick re-ran and confirmed the tests

PR #220, opened 2024-08-16
Update FB_MotionPneumaticActuator to include a flow meter

nrwslac:
## Description

-Adds `FB_FDQ_FlowMeter` to `FB_MotionPneumaticActuator`
## Motivation and Context

ST4K4 in TMO gets a flow meter.
## How Has This Been Tested?

Running on TMO motion, on local copy of common components. 
## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/Library.plcproj
+++ b/lcls-twincat-motion/Library/Library.plcproj
@@ -610,6 +610,10 @@
       <DefaultResolution>LCLS General, * (SLAC)</DefaultResolution>
       <Namespace>LCLS_General</Namespace>
     </PlaceholderReference>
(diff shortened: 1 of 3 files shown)


ZLLentz:
You should note that this FB is for general pneumatic motion, not just for the TMO stopper, though it seems like this has only been used for stoppers in practice:
https://github.com/search?q=org%3Apcdshub%20FB_MotionPneumaticActuator&type=code
Do all the cases in the above search also have flow meters added? (ST3K2, ST5K4). Are there any consequences to having the extra PVs, code, and link points available for instances without flow meters?

nrwslac:
On review, not all actuators utilizing FB_MotionPneumaticActuator will use an FDQ flow meter.

PR #221, opened 2024-10-29
Add motion virtual frame

NSLentz:
## Description

- Added a function block called FB_MotionVirtualFrame which takes:
  - 3 ST_MotionStages representing real axes that are orthogonal to each other and in a base frame of reference.
  - And 3 ST_MotionStages representing virtual axes that are orthogonal to each other and in a new frame of reference.
- Using these motion stages, it computes the transformation from a base frame of reference to a new frame of reference using 3 euler angles and a rotation order as inputs. By default, the order and angles are interpreted as going from the base to the virtual frame. This can be switched to go from virtual to base frame of reference.
 - This function block has two modes of operation. If enabled, this function block will automatically handle commanding the real axes to move when the virtual axes are commanded to move. If not enabled, the function block acts as a calculator, outputing the results of calculating motions in a new frame of reference, but not changing any setpoints or commanding any moves.

## Motivation and Context

- The MRCO Sample paddle and Gas Needle axes were designed and installed with their motor orientations not aligned with the xray beam axes.
- This made it difficult to move the targets and align them with the beam as the operator would need to move three different motors in stage coordinates in order to achieve a motion along one axis in beam coordinates.
- This function block provides an interface for the operator to move 3 Axis stages along arbitrary coordinate systems by computing the position, velocity, and accelerations required on each axis in order to get them to move in a coordinated fashion to achieve single axis motion in a different coordinate system.
- Measurements of the euler angles required to shift from one coordinate system to another can be obtained from a CAD model, and then verified with actual motion.
- This function block also has applications in areas that require extremely precise motions, as it provides a simple interface to correct for subtle angular deviations between the axes and the beam direction. For example, this would allow something like the roadrunner fast sample rasterization system to coordinate the motion of three separate axes to correct for small angular offsets of the sample holder as it moves into the path of the beam.

## How Has This Been Tested?

- Unit tests were created to test the new functionality. All unit tests are passing.
- Additionally, this system was tested live on the MRCO Sample Paddle 3 Axis system.
- Using a reference laser and a camera oriented facing down the path of the beam, motions were commanded along the X, Y, and Z axes in beam coordinates.
- The reference laser was observed to hit the sample paddle. Then when commanding the axes to move in beam coordinates, we could see if laser translated across the sample paddle in pure X and Z motions.
- We found that the X and Y motions were very close to pure motions in beam coordinates.
- Then we switched to a different camera that had an isometric view of the reference laser hitting the sample paddle.
- We commanded a pure Z motion in beam coordinates, and could see the laser point staying in the same relative spot on the sample paddle as it moved. This made it clear that the sample paddle was indeed moving in a pure Z motion, because otherwise, we would have seen the laser position deflect across the face of the sample paddle.

- The code was written in a verbose manner in an attempt to make it understandable and readable. Functional comments were written to describe the behavior of different inputs.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/Tests/FB_MotionVirtualFrame_Test.TcPOU
+++ b/lcls-twincat-motion/Library/Tests/FB_MotionVirtualFrame_Test.TcPOU
@@ -5,7 +5,9 @@
 FUNCTION_BLOCK FB_MotionVirtualFrame_Test EXTENDS TcUnit.FB_TestSuite
 VAR
     nTestIDAssigner : INT;
-    nTestID : INT;
+    nTestID : INT := 1;
+    tonTimeout : TON;
+    tonStart : TON;
 
     stMotionStageRx : ST_MotionStage;
     stMotionStageRy : ST_MotionStage;
@@ -26,6 +28,16 @@ VAR
     nPositionCountsVx AT %Q* : UDINT;
     nPositionCountsVy AT %Q* : UDINT;
     nPositionCountsVz AT %Q* : UDINT;
+
+    fbMotorTestSuite : FB_MotorTestSuite;
+    fbMotionStageSetAndMoveHelper : FB_MotionStageSetAndMoveHelper;
+    fbMotionStageSetHelperRx : FB_MotionStageSetHelper;
+    fbMotionStageSetHelperRy : FB_MotionStageSetHelper;
+    fbMotionStageSetHelperRz : FB_MotionStageSetHelper;
+    fbMotionStageSetHelperVx : FB_MotionStageSetHelper;
+    fbMotionStageSetHelperVy : FB_MotionStageSetHelper;
+    fbMotionStageSetHelperVz : FB_MotionStageSetHelper;
+    fbMotionVirtualFrame : FB_MotionVirtualFrame;
 END_VAR
 
 ]]></Declaration>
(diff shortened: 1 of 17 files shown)


lcls-twincat-motion/Library/Tests/FB_MotionVirtualFrame_Test.TcPOU
```diff
@@ -0,0 +1,1135 @@
...
+
+TestEnabled000090XYZMoveVxExpectNoErrorsCorrectPosVelAcc();
+
+TestEnabled009090XYZMoveVxExpectNoErrorsCorrectPosVelAcc();
+
+TestEnabled909090XYZMoveVxExpectNoErrorsCorrectPosVelAcc();
+
+TestEnabledN22N21P4XYZMoveVxVyVzExpectNoErrorsCorrectPosVelAcc();
+
+TestEnabledN22N21P4XYZMoveVyExpectNoErrorsCorrectPosVelAcc();
```
ZLLentz, line 60: I don't think these quite follow your own guidelines for unit test naming, unless I'm misunderstanding.

ZLLentz (approved):
There's a lot of code here. I read through looking for odd behavior and red flag practices but I didn't see any. It'll take way too much time to dive any deeper and this seemed to work in prod.

I noted two things that stood out to me.

lcls-twincat-motion/Library/Tests/PRG_TEST.TcPOU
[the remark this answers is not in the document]
NSLentz: I rearranged it to match the alphabetical order presented in the file tree to make it easier to carry out a visual confirmation that all test function blocks are actually being instantiated in the test POU.

lcls-twincat-motion/Library/Tests/FB_MotionVirtualFrame_Test.TcPOU
NSLentz: They are more complicated than I would like but that is due to the motion library being structured in such a way that I cannot test new functions that rely on the motion stage code without doing larger and more complicated unit tests that are difficult to describe.

NSLentz: I think a future improvement to the motion library would be to break the function blocks used within an FB_MotionStage into a set of interfaces that are provided at instantiation of an FB_MotionStage. That way I could mock it out in smaller pieces and even mock an NC axis so I don't have to use simulation axes anymore.

PR #222, opened 2024-12-20
Retrying bptm when enable is toggled on

KaushikMalapati:
## Description

I  added an R_TRIG to FB_MotionBP to detect rising edges in bEnable and passes that as the input to bRetry. 

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-6609

Tong noticed that if IM5K4 was in a state that had a preemptive request, like when it's at the YAG1 position, if you disable the arbiter (IM5K4:PPM:MMS:STATE:PMPS:ARB:ENABLE_RBV), the preemptive request goes away like it should. However, if you reenable the arbiter without changing the position, the preemptive request does not come back unless you move to a different state first. 

## How Has This Been Tested?

Not yet tested

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MotionBPTM.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MotionBPTM.TcPOU
@@ -36,6 +36,8 @@ VAR_INPUT
     bMoveOnArbiterTimeout: BOOL := TRUE;
     // Set to TRUE when it is safe to reset the BPTM timeout fast fault, to reset it early.
     bResetBPTMTimeout: BOOL;
+	// Set to TRUE to make the BPTM retry
+	bRetry: BOOL;
 END_VAR
 VAR_OUTPUT
     // This becomes TRUE when the motors are allowed to move to their destinations.
@@ -118,6 +120,7 @@ bptm(
     i_stRequestedAssertion:=stGoalParams.stBeamParams,
     i_xDoneMoving:=bDoneMoving AND bAtState,
     stCurrentBeamParameters:=PMPS_GVL.stCurrentBeamParameters,
+	bRetry:=bRetry,
     q_xTransitionAuthorized=>bInternalAuth,
     bDone=>bDone,
 );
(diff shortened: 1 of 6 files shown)


lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MotionBPTM.TcPOU
ZLLentz, line 1: Link to BPTM API for reviewers: https://github.com/pcdshub/lcls-twincat-pmps/blob/master/lcls-twincat-pmps/PMPS/MajorComponents/BPTM/BeamParameterTransitionManager.TcPOU

I wonder if we should be using this elsewhere too, like whenever a stopper is pulled to resolve BPTM timeouts.

ZLLentz (commented):
I think the concept is a good idea, hopefully it's possible to test before merging

NSLentz:
I think what I need to evaluate this change is the following:
(1) Unit test needs to be written which replicates the problem.
(2) Implement this fix
(3) Demonstrate that the problem is fixed by the implementation
(4) Show that all other unit tests still pass

NSLentz:
@KaushikMalapati have you been able to test on a project yet? The way to do this would be to install your version on the library on plcprog as version 0.0.0. Then explicitly select version 0.0.0 in the test project as the placeholder. Then run your live tests.

lcls-twincat-motion/Library/Tests/FB_PositionStatePMPSND_Test.TcPOU
```diff
@@ -864,7 +870,7 @@ CASE nState OF
         IF fb_Move1D.stPlcToEpics.bDone THEN
             AssertFalse(
                 fbArbiter1D.CheckRequestInPool(astBeam[E_TestStates.TARGET1].nRequestAssertionID),
-                'Destination bp should have been in the arbiter',
+                'Destination bp should have been in the arbiter2',
```
NSLentz, line 873: Should this say "should not have been in the arbiter2"?

NSLentz (dismissed):
I think this looks good. nTestID is only needed for tests that you want to run multiple times with different parameters each time, but fine to have regardless. It looks correct for the WaitForBP state to occur after the request has been added to the pool, based on my dive into the PMPS library, so I don't see anything wrong with this.

NSLentz (approved):
Approved, thanks!

PR #223, opened 2025-01-21
add nRawEncoderDINT field variable  for EL5072 LVDT terminal position…

baljamal:
… reading

## Description
lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT: add nRawEncoderDINT field variable  for EL5072 LVDT terminal position reading
## Motivation and Context
TMO DREAM KB slits has 2xEL5072 terminals to interface 4xLVDT position sensors.
EL5072 position INPUT variable has DINT type (raw position reading in nm).

## Pre-merge checklist


ZLLentz (commented):
This looks correct to me. This doesn't have a functional use inside the PLC code as far as I know but is used to display an accurate encoder count in the UI.

baljamal:
> This looks correct to me. This doesn't have a functional use inside the PLC code as far as I know but is used to display an accurate encoder count in the UI.

yes, you are right

PR #225, opened 2025-01-31
Backlash compensation support for open and closed loop motion.

yannS2016:
## Description
The new feature adds support for backlash correction in the LCLS TwinCAT motion library for assisting in more precise positioning for both closed and open loop systems.

- Add EPICS support in ST_MotionStage for Backlash compensation user interface
- Created an FB_MotionBacklashCompensation wrapping MC_BacklashCompensation to implement compensation functionalities
- Add call for FB_MotionBacklashCompensation and expanded error handling in FB_MotionStage

## Motivation and Context

The feature is required to fix a vital need for more precise positioning of open and closed loop systems facing motion backlashes.
https://jira.slac.stanford.edu/browse/ECS-6374

## How Has This Been Tested?

The Backlash compensation algorithm has been tested for:

- Disabling backlash compensation when homing an axis.
- Updating the compensation value and setting the direction based on the sign of the user interface variable ( EPICS interface)
- Enabling compensation when a move command is registered and updating user interface variables.
- Applying compensation in the direction of the backlash
- The feature has been tested on a LAB setup organized as running from a local machine running the latest TWinCAT ( 4026.13) with dependant LCLS libraries compiled under the same version of TwinCAT.

**The testing setup consisted of:**

Beckhoff EL7047 drive and EL5101 incremental encoder attached to a stepper motor served as the stage.
An NC axis was added to the library under test and linked to the staged hardware
A sample PLC program leveraging the LCLS TwinCAT motion control library resources was used to test the feature

**NB:** _The NC axis configuration parameter "Position Correction" must be set "TRUE" before activating the configurations._

**The Validation of the compensation feature consisted of:**

- Homing the axis and verifying that the move was done without compensation
- Setting a compensation value and observing the direction of backlash compensation correction
- Registering a move in one direction and reversing to observe compensation being applied.

This observation was made by capturing the encoder count at a given position A, then at position B, and reversing to position C without any compensation being applied. Then going through the same process from the same position A with backlash compensation enabled, a user-defined compensation value set, thus observing the final count of the encoder at position C.

The results showed that the compensated encoder count at C is that of the uncompensated one minus the compensation value in encoder counts. this held for both backlash directions.

The feature required that some code be added to the MotionStage ST and B. The backlash compensation feature implementation mainly reads in two inputs from variables in FB_MotionStage and expands its error handling.

**NB:** _Unit tests for this feature have not been developed yet. So testing only deals with what's existing including the added support for backlash compensation._

the FB_MotionBacklashCompensation FB carries light documentation of the feature implemented and the required initial conditions.
## Pre-merge checklist
- [X] Code works interactively
- [X] Test suite passes locally
- [X] Code contains descriptive comments
- [X] Libraries are set to ``Always Newest`` version (``Library, *``)
- [X] Committed with ``pre-commit`` or ran ``pre-commit run --all-files``


ZLLentz:
In the future it's ok to keep the same branch and PR and push new commits to it, they will update in place.

ZLLentz:
One question about the approach before I dive into the code:

> Disabling backlash compensation when homing an axis

Why do we need to disable the backlash compensation when homing an axis?

I'll read through the code now, thank you for the detailed description of your changes and testing setup!

lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT
```diff
@@ -218,6 +218,43 @@ STRUCT
...
+    {attribute 'pytmc' := '
+        pv: PLC:fBacklashCompensation
+        io: io
+        field: DESC Backlash compensation
+    '}
+    fBacklashCompensation: LREAL;
+
+    // Current Backlash compensation value ?
+    {attribute 'pytmc' := '
+        pv: PLC:fCurrentBacklashCompensation
```
ZLLentz, line 253: You're going to want to keep these short because there's a limit on total PV character count in our EPICS setup. This string `PLC:fCurrentBacklashCompensation` will end up being part of a full motor name which might be e.g.

`IM3L0:PPM:MMS:PLC:fCurrentBacklashCompensation`

So depending on what the base PV is this could get pretty close to the character limit (which is 40? 60? I can't remember the exact number). I suggest keeping these pv suffixes down to about the same length as other elements in the struct.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionBacklashCompensation.TcPOU
```diff
@@ -0,0 +1,46 @@
...
+VAR_INPUT
+    bHoming: BOOL;
+    bMoving: BOOL;
+END_VAR
+VAR_OUTPUT
+    bEnabled: BOOL;
+    bBusy: BOOL;
+    bError: BOOL;
+    nErrorID: UDINT;
+END_VAR
```
ZLLentz, line 17: This function block needs some documentation. What are the inputs and outputs supposed to be?

Also, `bEnabled` and `bBusy` are unused.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionBacklashCompensation.TcPOU
```diff
@@ -0,0 +1,46 @@
...
+VAR
+    fbBacklashCompensation: MC_BacklashCompensation;
+    bPrevCompensation : LREAL := 0.0;
```
ZLLentz, line 20: This appears to use hungarian notation, but it's using a `b` for boolean instead of e.g. an `f` or something else to represent the float/lreal

lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT
```diff
@@ -218,6 +218,43 @@ STRUCT
...
+    (* Backlash compensation*)
+    // Enabled axis backlash compensation
+    {attribute 'pytmc' := '
+        pv: PLC:bBacklashEn
+        io: io
+        field: ZNAM FALSE
+        field: ONAM TRUE
+        field: DESC Enable Backlash compensation
+    '}
+    bUserBacklashEn: BOOL;
```
ZLLentz, line 231: Are there any design considerations for expecting the user to configure this in e.g. EPICS or in the PLC program vs in the NC?

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
```diff
@@ -170,6 +176,11 @@ fbMotionHome(
 stMotionStage.Axis.ReadStatus();
 
 // Check for a new error
+IF NOT stMotionStage.bError THEN
+    stMotionStage.bError := fbBacklashCompensation.bError;
+    stMotionStage.nErrorId := fbBacklashCompensation.nErrorId;
+END_IF
```
ZLLentz, line 182: It seems to me like the backlash compensation errors now have priority over all other errors for user reporting, since it comes first in the code here! Is this what we want?

I remember Nick had some thoughts about better way of propagating errors so this might not be so important.

ZLLentz (commented):
Overall I think this is great!

I have some small questions and nitpicks, some of them are for discussion more than as suggestions

lcls-twincat-motion/Library/POUs/Motion/FB_MotionBacklashCompensation.TcPOU
yannS2016: Alright, will add some extra info!

lcls-twincat-motion/Library/POUs/Motion/FB_MotionBacklashCompensation.TcPOU
yannS2016: oh sure!

lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT
yannS2016: The NC configurations need to set "Position Correction" for the subsequent logic to be used. there is a logic that sets the MC block enable input based on this user enabled and some other conditions.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
yannS2016: Thanks for pointing this out. should we queue errors before reporting?

lcls-twincat-motion/Library/DUTs/ST_MotionStage.TcDUT
yannS2016: good catch!

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
NSLentz: @ZLLentz discussed this with @yannS2016. It might be time for us to implement an error queuing system. I've been running into issues with only the last error being displayed. Yann will look into the best way to handle this.

NSLentz (dismissed):
I think this looks great, I think this is OK to merge as is. But our next step should be to work out how we can handle errors better. Via queuing system or priority system or combination thereof.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
yannS2016: Alright. ill report back when i figure something out. thanks for the feedback.

ZLLentz (approved):
Thank you for the updates and thoughtful responses to my review questions!

yannS2016:
I added a unit test for the backlash compensation feature.

NSLentz (approved):
Looks good to me.

PR #226, opened 2025-02-19
Reset Drive Enable Mode after mode transition

yannS2016:
## Description

Reset Drive Power after a mode transition of "EnableMode" from 'ALWAYS' to 'DURING_MOTION' 
## Motivation and Context

Bug fix: The motion library does not remove power on the drive when mode switches from ALWAYS to 'DURING_MOTION'. this start motion in an unsafe condition as 'DURING_MOTION' should first enable power to drive and then execute the motion request.
## How Has This Been Tested?

Code Tested on TwinCAT 4026.11 under TcBSD PLC running TwinCAT 4026 connected to a Renishaw absolute stage with an EL7031 drive. Library Units Test were successfully run. The feature functionality tests consisted of setting the enable mode to 'ALWAYS' verifying that power was enabled in all directions then resetting the enable mode to 'DURING_MOTION' and observed that power was removed to the drive in all direction. Move request in this mode would then follow the expected sequence of actions.

Descriptive comments were added above implementation line in the modified FBs.
## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
@@ -35,6 +35,7 @@ VAR
     bMoveCmd: BOOL;
     rtMoveCmdShortcut: R_TRIG;
     rtHomeCmdShortcut: R_TRIG;
+    rtEnableMode: R_TRIG;
 END_VAR
 ]]></Declaration>
     <Implementation>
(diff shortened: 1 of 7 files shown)


lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
ZLLentz, line 1: I suspect this fixes the issue :+1:

Is this worth adding a unit test for? It seems pretty simple to test compared to some of the other things we try to do in this library.

yannS2016: No, didn't add unit test, just meant that i rerun the normal unittest to ensure the feature did not break the library.

NSLentz (changes requested):
Yann, could you add a unit test for this? I think it would be good to capture this check in a test, especially for when we want to update the library in the future. This will indicate to us something to check for.

yannS2016:
> Yann, could you add a unit test for this? I think it would be good to capture this check in a test, especially for when we want to update the library in the future. This will indicate to us something to check for.

@NSLentz  added a unit test.

lcls-twincat-motion/Library/POUs/Motion/FB_MotionStage.TcPOU
yannS2016: Unit test added

NSLentz (approved):
looks good, thank you

PR #227, opened 2025-03-05
Warning Message for unset bHomed Flag for a Home mode defined stage

yannS2016:
## Description

Implement warning message when the axis homing mode is assigned and the bHomed flag is not set before attempting to move the axis
## Motivation and Context

EPICS side complains while an axis with homing mode defined, but bHomed is not set. in other words, no homing routine was executed for this axis before an absolute move was requested.
## How Has This Been Tested?

Code Tested on TwinCAT 4026.11 running in Local Mode under VM.  A warning message in printed in the error message window each time an absolute move is performed on the axis while bHomed is unset to notify the operator. this notification will be reset once the axis is homed.
bhomed

Descriptive comments were added above the implementation line in the modified FBs.
## Pre-merge checklist


NSLentz (approved):
This looks good to me. Simple change but I think it will help a lot.

PR #231, opened 2025-04-22
Debouncer autoreset and counter max limit

KaushikMalapati:
## Description

- Making the debouncer fast fault auto-reset
- nTripCount now has a maximum value of 5, so that it does not grow larger than that and delay beam while it ticks down

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-7551

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MiscStatesErrorFFO.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MiscStatesErrorFFO.TcPOU
@@ -97,7 +97,7 @@ ffUnknown(
 ftTripCount(CLK:=ffBeamParamsOk.i_xOK);
 
 IF ftTripCount.Q THEN
-    nTripCount := nTripCount + 1;
+    nTripCount := MAX(nTripCount + 1, 5);
 END_IF
 
 tonTripCount(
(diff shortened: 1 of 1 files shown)


ZLLentz (approved):
I think this is pretty clearly correct. I'm going to add motion-sme as a reviewer since it's the motion repo, we should get their eyes on this too.

ZLLentz:
I'm not twincat-reviewers or motion-sme but consider my review as being for the pmps-sme rep

NSLentz (approved):
I agree, this is clearly a good change. Approved.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MiscStatesErrorFFO.TcPOU
```diff
@@ -97,7 +97,7 @@ ffUnknown(
 ftTripCount(CLK:=ffBeamParamsOk.i_xOK);
 
 IF ftTripCount.Q THEN
-    nTripCount := nTripCount + 1;
+    nTripCount := MAX(nTripCount + 1, 5);
```
ZLLentz, line 100: In retrospect, I think we might need to put this at a higher number than 5, else the debouncer is itself bouncy because it trips exactly at 5. Any thoughts?

ZLLentz: Alternatively, we can make it reset automatically at 0 instead of at 4

ZLLentz: Maybe I'll make a small PR to make it reset only at zero

PR #232, opened 2025-04-23
Vacuum Library 4026 Migration: Migrate LVDT part to the motion library

yannS2016:
## Description

4026 Migration: Move the LVDT part  of the Vacuum Library to the motion library
## Motivation and Context

Upgrade the architecture for the vacuum Library. The LVDT part of the vacuum Library handles motion feedback. For the 4026 Migration and uniformity, it was agreed to move that section to the motion library.
## How Has This Been Tested?

The LVDT device used in the vacuum subsystems is the Serial device, thus it will require communication with real hardware for unittesting.

No modification was made to the imported code.
## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/DUTs/LVDT/ST_SerLVC4000.TcDUT
+++ b/lcls-twincat-motion/Library/DUTs/LVDT/ST_SerLVC4000.TcDUT
@@ -1,0 +1,59 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<TcPlcObject Version="1.1.0.1" ProductVersion="3.1.4022.16">
+  <DUT Name="ST_SerLVC4000" Id="{1ecc4b20-fded-4e3a-82c1-cf810d73e466}">
+    <Declaration><![CDATA[TYPE ST_SerLVC4000 :
+STRUCT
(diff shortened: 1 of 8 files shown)


ZLLentz:
There's clearly nothing in the code here about motion feedback. What does this code actually do? When would someone use these? Why did I get away with not using these when I implemented LVDT control in the EBD with the LVC-4000s? Why are there so many function blocks whose comments refer to e.g. pumps?

yannS2016:
> There's clearly nothing in the code here about motion feedback. What does this code actually do? When would someone use these? Why did I get away with not using these when I implemented LVDT control in the EBD with the LVC-4000s? Why are there so many function blocks whose comments refer to e.g. pumps?

Sure, after digging a bit regarding these LVC4000 devices it appears that it's been used for: https://confluence.slac.stanford.edu/spaces/L2SI/pages/263755992/HXR+K-Monochromator+Motion. a github project using the device: https://github.com/pcdshub/lcls-plc-lfe-motion-kmono. Besides this, I don't see where this serial interface is being used too as it pertains to motion. I'll keep it in the vacuum lib for now and wait for @jyin999 to clarify this.

PR #234, opened 2025-05-02
Fixed virtual frame bug and tests that were not finishing

NSLentz:
## Description

- Fixed the virtual frame function block not commanding moved bug. 
- Also found that tests were not finishing. 
- Traced the issue to backlash compensation and power disable on during motion tests. 
- Backlash compensation test axis reference was never linked to the axis.
- Added timeouts so that at least tests finish and we can see that backlash compensation test is failing.

## Motivation and Context

- Kaushik caught these issues when making a separate pull request.

## How Has This Been Tested?

- Ran the unit tests and now all pass except debouncer test that Kaushik is already fixing in another PR, and backlash compensation test doesn't pass. Will need to ask Yann to fix that.

## Pre-merge checklist


lcls-twincat-motion.sln
yannS2016, line 1: Nick, I've identified some issues in the state machine. You may just address the Virtual drive bug fix here; I'll deal with the fixes in another PR.

NSLentz: Understood, could you review the PR so that I can merge it?

yannS2016 (approved):
All good! Just remove the parts  test_finished timers  as discussed

PR #238, opened 2025-07-03
Running bptm timeout fault when arbiter is disabled and passing bEnable

KaushikMalapati:
## Description

- In FB_MotionBPTM, HandleTimeout() is executed whether bEnable is true or false, instead of only when bEnable is true. This should prevent the fault from being unclearable if bEnable becomes false
- ffBPTMTimeoutAndMove resets when bEnable is false
- ffBPTMTimeoutAndMove is OK if ffBPTMTimeoutAndMove is false

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-7791

## How Has This Been Tested?

Tested briefly on tmo-motion by setting setting bArbiterEnabled to true and false. I saw at least once that the fast fault was activated as soon as bEnable was set back to True (it was ok when bEnable was false). Maybe bptm.bBusy was true the entire time, so TON.Q stayed high? If that's the only issue, I think I could assign `TON.IN := bptm.bBusy AND bEnable` to reset the timer when disabling. I think combined with the fact that the bptm resets when bEnable has a rising edge, this would stop the fault from happening when bEnable has a rising edge. 

## Pre-merge checklist


NSLentz (approved):
looks good to me

PR #245, opened 2025-08-14
Add first cycle timer to read and update NC vars from PLC after autoboot

yannS2016:
## Description

context: https://jira.slac.stanford.edu/browse/ECS-8045
Add first cycle delay to read and update NC vars from PLC after autoboot
## Motivation and Context

Add first cycle delay to read and update NC vars from PLC after autoboot
## How Has This Been Tested?

image

## Pre-merge checklist


NSLentz (dismissed):
There is a slightly more robust way to ensure the NC is ready for communication and that is to test the axis id and wait until it is not equal to 0. This means that the NC is ready. However, what you have here will work anyways but perhaps in a future update you could make sure to use the axis id <> 0 test.

PR #247, opened 2025-10-30
feat: added prgma to enable autosave on selected PVs

jozamudi:
## Description

Added autosave to the following PVs using pass1.

ST_PositionState.fVelocity
ST_PositionState.fPosition
ST_MotionStage.bUserEnable
ST_MotionStage.fHomePosition
ST_MotionStage.bUserBacklashEn
ST_MotionStage.fBacklash

## Motivation and Context

https://jira.slac.stanford.edu/browse/ECS-9149

### Test procedure
I wrote a unit test that ... 
1. starts the IOC
2. clears the PVs 
3. writes some values into the PVs
4. restart the IOC
5. check if the written values are restored.

We found some interesting autosave behavior. If the PLC restarts while the IOC is rebooting the autosave_pass0 PVs will lose the autosaved value. 

## Pre-merge checklist


yannS2016 (commented):
I'd like to know whether a PLC program with hardcoded state fPosition/fVelocity via cyclic fbStateSetup calls wouldn't just overwrite anything from autosave.  If an IOC reboots while a PLC is active, wouldn't the PLC just assert the hardcoded ones a couple of cycles later? A rebooting PLC is not 'online', so the rebooting IOC is not able to establish and complete the correct boot-up sequence. The other boolean flags could be restored in pass0|pass1.

yannS2016:
> The behavior I saw with autosave, the PVs are only written once. If the PLC has a hard coded value the PLC would overwrite the value.

Depending on how Asyn reads methods in device support are implemented, with or without the hardware issues, init_record for output records tends to overwrite the autosave pass0 restore during Asyn's first hardware read(output records). Could you check if the input records get the restore values in both cases at pass0?

yannS2016 (commented):
Perhaps a second pair of eyes is needed here. Otherwise this LGTM.

NSLentz:
@yannS2016 why are we talking about pass0? The new pragmas as using pass1. We tested this and it works to restore the parameters. This is an improvement on the existing code I don't see why we would hold up this PR

@NSLentz  The PR mentioned pass0, which till now is the default restore method for PVs from this library. We were merely trying to figure out the observation made here. I would like to see a thorough review of how autosave is used overall for the motion library PVs and how pass1 is an improvement.

NSLentz (approved):
approved. I was present for testing

PR #250, opened 2026-02-24
Reduce FFO fatigue and simplify commissioning state positioners through single boolean ffo off using maintenance mode

NSLentz:
- I am making this PR to intentionally start debate on this topic.

## Description

- I am proposing this change in order to simplify the user experience during commissioning.
- The proposal is for Arbiter enable/disable to still do the same thing except turning a pmps state positioner into maintenance mode would now make only one ffo occur, and that is an ffo to state you are in maintenance mode. All other ffos in the pmps state positioner would be blocked.

## Motivation and Context

- I argue that the system as is too confusing from the user perspective. Even as SME, I can never remember what the maintenance mode and arbiter enable/disable are supposed to do. This change would simplify the user experience. I argue that we will have a better user experience if they can just take the system under maintenance level control from one boolean.
- Additionally, users already seem to expect this behavior anyways.

## How Has This Been Tested?

- This change has not yet been tested. I propose we debate the merits first before committing to implementing and testing the changes.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MotionBPTM.TcPOU
+++ b/lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MotionBPTM.TcPOU
@@ -96,7 +96,7 @@ bTransitionAuthorized := bInternalAuth OR (bArbiterTimeout AND bMoveOnArbiterTim
 
 // Trip the beam for BPTM timeouts if we want to move
 // Only reset at safe beam OR at no bptm errors (some other FF should catch additional issues)
-ffBPTMTimeoutAndMove.i_xOK := NOT (bArbiterTimeout AND bMoveOnArbiterTimeout AND bEnable);
+ffBPTMTimeoutAndMove.i_xOK := NOT (bArbiterTimeout AND bMoveOnArbiterTimeout) OR (NOT THIS^.bEnable);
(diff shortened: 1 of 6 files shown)


ZLLentz:
I agree with the general approach here- I might have a related thought about reducing the number of FFOs in general. But this is a good idea.

lcls-twincat-motion/Library/POUs/Motion/PMPS/FB_MotionBPTM.TcPOU
```diff
@@ -96,7 +96,7 @@ bTransitionAuthorized := bInternalAuth OR (bArbiterTimeout AND bMoveOnArbiterTim
 
 // Trip the beam for BPTM timeouts if we want to move
 // Only reset at safe beam OR at no bptm errors (some other FF should catch additional issues)
-ffBPTMTimeoutAndMove.i_xOK := NOT (bArbiterTimeout AND bMoveOnArbiterTimeout AND bEnable);
+ffBPTMTimeoutAndMove.i_xOK := NOT (bArbiterTimeout AND bMoveOnArbiterTimeout) OR (NOT THIS^.bEnable);
```
ZLLentz, line 99: I'm rusty on my Beckhoff OOP- what do these changes do?

PR #252, opened 2026-06-03
Replace E_DS402OpMode with E_MCS2OpMode and fix bParamsSet latch

yannS2016:
## Description

Renames the DS402 operation-mode enum from `E_DS402OpMode` to `E_MCS2OpMode` with the new enum's values (CSP/HOME/STEP/CALIBRATE) better reflecting the MCS2-specific use. Promotes `eRequestedManualMode` on `FB_MCS2DriveManager` to `VAR_OUTPUT` and wires it into `FB_MotionStageMCS2` as `nCommandLocal`, whose CASE statement now keys off `E_MCS2OpMode` values directly (instead of `E_EpicsMotorCmd`); `nCommandLocal` is no longer PERSISTENT. Fixes `bParamsSet` in `FB_WriteChanParameters`: the prior `R=` cleared the flag when any writer went busy but never restored it, so a single parameter write left `bParamsSet` permanently FALSE.

## Motivation and Context

DREAM MCS2 motion control.

## How Has This Been Tested?

Tested on MCS2 driving a couple of stick-slip stages. Verified mode change on demand, mode display via the new output PV, and that `bParamsSet` correctly returns to TRUE after writer activity completes.

Inline comments next to the updated `bParamsSet` logic in `FB_WriteChanParameters` describe the corrected real-time tracking behavior and call out the prior bug.

## Pre-merge checklist

### DIFF
--- a/lcls-twincat-motion/Library/DUTs/DS402/E_MCS2OpMode.TcDUT
+++ b/lcls-twincat-motion/Library/DUTs/DS402/E_MCS2OpMode.TcDUT
@@ -1,10 +1,10 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <TcPlcObject Version="1.1.0.1">
-  <DUT Name="E_DS402OpMode" Id="{6b34f554-8c74-4ff3-b4ed-290b296a94b3}">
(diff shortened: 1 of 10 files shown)


NSLentz (approved):
lgtm

